PR #43, opened 2014-12-22
Improve configuration and servlets

alechenninger:
A lot of things bothered me about how the clients were configured, so abstracted much of the coupling out to a separate class, `ProperitesLightblueClientConfiguration`.

Also, did some other cleanup around the use of configuration.

Let me know if there are any questions.


alechenninger:
Hold off on review, working on doing things a little more "proper" with Java CDI injection.

alechenninger:
With the latest changes, it is now possible to use a servlet for each data and metadata (as is most intuitive), and they will share the same Apache `CloseableHttpClient` so the same connection pool can be used.

Will add Javadoc, but for example:

web.xml
```xml
  <servlet>
    <servlet-name>lightblue-data</servlet-name>
    <servlet-class>com.redhat.lightblue.client.http.servlet.LightblueDataProxyServlet</servlet-class>
  </servlet>
  <servlet-mapping>
    <servlet-name>lightblue-data</servlet-name>
    <url-pattern>/lightblue/data/*</url-pattern>
  </servlet-mapping>

  <servlet>
    <servlet-name>lightblue-metadata</servlet-name>
    <servlet-class>com.redhat.lightblue.client.http.servlet.LightblueMetadataProxyServlet</servlet-class>
  </servlet>
  <servlet-mapping>
    <servlet-name>lightblue-metadata</servlet-name>
    <url-pattern>/lightblue/metadata/*</url-pattern>
  </servlet-mapping>
```

To allow the `CloseableHttpClient` to be injected you need a `beans.xml` file in the WEB-INF directory (it can be empty) and a simple producer/disposer:

```java
public class HttpClientProducer {
    @Produces @ApplicationScoped
    public CloseableHttpClient getClient() {
        // Defaults look for the lightblue-client.properties file
        return LightblueHttpClientBuilder.buildWithDefaults();
    }
    
    public void closeClient(@Disposes CloseableHttpClient client) throws IOException {
        client.close();
    }
}
```

Unfortunately the producer can't be included inside the lightblue-client dependency because of how bean archives work with Java CDI. However as you can see it is trivial to write. Writing your own also allows you to configure it however you wish.

alechenninger:
@derek63 @jewzaam This is officially ready for review. Thanks for your patience!

Summary of changes (previous comments are out of date at this point):
- Removed lightblue-client.properties lookup code from several places which duplicated functionality, and move this to a [PropertiesLightblueClientConfiguration](https://github.com/lightblue-platform/lightblue-client/pull/43/files#diff-2) class which has factory methods to create configuration instances from properties files or objects. For how this is used compared to before see [this as an example](https://github.com/lightblue-platform/lightblue-client/pull/43/files#diff-7978d9eb5bf965397e6d1db133a9c02aL44).
- Allow injecting a different ObjectMapper into LightblueHttpClient, and use shared static implementation by default (this is recommend by jackson library creators). See [here](https://github.com/lightblue-platform/lightblue-client/pull/43/files#diff-7978d9eb5bf965397e6d1db133a9c02aR35) and [here](https://github.com/lightblue-platform/lightblue-client/pull/43/files#diff-7978d9eb5bf965397e6d1db133a9c02aR75)
- Refactored out [factory classes for Apache HttpClients](https://github.com/lightblue-platform/lightblue-client/pull/43/files#diff-9) and [SSLConnectionSocketFactories](https://github.com/lightblue-platform/lightblue-client/pull/43/files#diff-13) from the HttpClientAuth implementations. This is so it is easier to create custom HttpClients for injecting into servlets (see below) or perhaps the LightblueHttpClient itself in the future. This also deprecates the HttpClientAuth interface and implementations, because the factory classes do the same thing but a little more explicit I think. The HttpClientAuth implementations did not allow for further customization of the client (such as a custom connection pool). They have been [refactored to use the new factories](https://github.com/lightblue-platform/lightblue-client/pull/43/files#diff-c2147df53c0c67481d7b909ca7295540L72) instead where appropriate.
- Added new servlets that work with the [newer lightblue-client.properties configuration scheme](https://github.com/lightblue-platform/lightblue-puppet/blob/master/templates/client/lightblue-client.properties.erb). The [old LightblueProxyServlet is now deprecated](https://github.com/lightblue-platform/lightblue-client/pull/43/files#diff-3d241f0bd8fa39c7c307334ef8f3ec4bR30), because it uses a `servletURI` property, which is no longer laid down by puppet. The new properties files instead have `dataServiceURI` and a `metadataServiceURI` properties (this is old news). So, I made new servlets, [one for data](https://github.com/lightblue-platform/lightblue-client/pull/43/files#diff-15), and [one for metadata](https://github.com/lightblue-platform/lightblue-client/pull/43/files#diff-16), that will proxy for each service respectively, using the right service URI defined in the configuration. The new servlets also have a few other improvements:
    - [EE CDI Injectable HttpClient dependency](https://github.com/lightblue-platform/lightblue-client/pull/43/files#diff-d455d921198cedf943fe17a10beb5972R47) so that a single HttpClient can be shared across multiple servlets. It is thread safe and backed by a thread pool so this is a recommended practice. This also allows a mock client to be injected in unit tests. The tradeoff is consumers have to write a trivially small producer/disposer class, add a beans.xml, and deploy to a container that supports CDI (I think it is standard with EE6). An example producer/disposer class is given in the Javadoc.
    - [Configurability through servlet init parameters](https://github.com/lightblue-platform/lightblue-client/pull/43/files#diff-17). Perhaps not a big deal but it was trivial to add.
    - Aside from reusing the http client, I think the servlets might be more efficient because they don't load the whole response into memory, but instead [buffer it using streams](https://github.com/lightblue-platform/lightblue-client/pull/43/files#diff-d455d921198cedf943fe17a10beb5972R96).
- Unit tests for the new servlets

Issue #42, opened 2014-12-22
Inject an HttpClient instead of new-ing it up internally inside LightblueHttpClient


alechenninger:
From https://hc.apache.org/httpcomponents-client-ga/tutorial/html/connmgmt.html

> 2.1. Connection persistence
The process of establishing a connection from one host to another is quite complex and involves multiple packet exchanges between two endpoints, which can be quite time consuming. The overhead of connection handshaking can be significant, especially for small HTTP messages. One can achieve a much higher data throughput if open connections can be re-used to execute multiple requests.

Sure, it makes it easy for clients to use because they don't have to worry about handling an HttpClient in their code, but this seems lazy to me. There should be a better way for clients to manage the life cycle of an HttpClient in various service and web frameworks that they use, rather than have the service client make a new connection for every service call. Currently the lightblue client gives no such option; it must manage the HttpClient, and gives no interface for a consumer to close it at some point, so it must close it after every request.

alechenninger:
Fixing this is fairly straightforward though not trivial, and in doing so would close #50 as well.

Currently, `LightblueClient` news up its own `HttpClient` dependency by way of `HttpClientNoAuth` and `HttpClientAuth` classes, which are really just factories for Apache `HttpClient`s. This has a number of problems, which is in part why those classes are deprecated. The most egregious of these is #42 , in addition to this class being non-unit-testable, because it will always try to make real HTTP requests.

The solution is the dependency injection (aka inversion of control) pattern. This means that `LightblueClient` should accept an `HttpClient` in it's constructor, instead of "controlling" the implementation of that dependency itself.

However, one step further would be to avoid coupling to Apache's `HttpClient` interface at all, and instead provide our own simple interface. Another option would be to use [Google's http client library][1], which is exactly that: an abstraction around various possible HTTP client implementations, which may be Apache, vanilla java.net, or anything else. Of course, there is some concern around coupling to Google's API's themselves, so it seems as though the best compromise is to roll our own (though there can still be an implementation that uses Google's APIs, which would probably be a good idea to leverage that OSS work that is already done.. I'm reminded of effective java 47: Know and use the libraries).

Work that would need to be done:
- [ ] Design API for http clients. Will need to be at least three interfaces: Request, Response, and the Client itself. (Names TBD)
[...]

PR #47, opened 2015-01-19
Use Basic connection manager instead of Pooling

alechenninger:
Because a new client is instantiated for each request, there is no need for a connection pool behind each client.

This will help #42 's reported performance implications


jewzaam:
I don't know the client code and didn't have time to dig in.. does this change not require any additional tests?

derek63:
@alechenninger This PR looks good and I am ready to merge the code, but it seems there are some merge conflicts.  Can you resolve those when you have the chance?

luan-cestari:
For me it is OK to merge, I'm just unsure about the pooling change made, I understood that it is using the BasicHttpClientConnectionManager instead teh custom one (and some code were refactored), right?

alechenninger:
Basically this makes the LightblueClient instance use a single threaded connection manager instead of a multithreaded one. The client instantiates and closes an entirely new client every single request, and prior to this PR it is starting up and spinning down a threadpool because that's what the default HttpClient does.

In the future the LightblueClient _really needs_ to have an injected HttpClient instead of newing it up itself. In my humble opinion I think this is a very fundamental design flaw. Based on the conversations I've had with @kahoali @jblashka and @derek63 we intend to make our own kind of "HttpClient" interface which can be implemented by plain old java.net or Apache's client or whatever. Until then, we can at least make the HttpClient implementation that the LightblueClient is newing-up internally not also new up a thread pool only to tear it down immediately.

luan-cestari:
@alechenninger Cool. I will merge the changes. Could you open an issue for the feature that you mentioned (you probably have much more context than I to put on the description of the issue)? Also you mentioned injected, did you think we could use something like CDI (or Weld or Guice or etc)?

alechenninger:
I don't mean any framework specific injection, I just mean dependency injection as the general technique of passing a class's dependencies instead of a class newing them up itself, which can be implemented with just plain Java via constructor injection (which is what I would propose in this case--consumers could wire this up with a framework if they desired.)

alechenninger:
#42 is really about the dependency injection issue, though I did not fully recognize it when the issue was initially created. I've updated it.

PR #48, opened 2015-01-20
Travis update

luan-cestari:
The project is now following the same configuration as the https://github.com/lightblue-platform/lightblue-rest (but with a different encrypted key/pass to deploy to remote repository).


jewzaam:
@luan-cestari can't be merged

PR #55, opened 2015-02-25
Include more information in exception

dcrissman:
Just look at the second commit, the first is just changing tabs to spaces.

More information is needed when debugging client issues. This PR will provide more information.

### DIFF
--- a/http/src/main/java/com/redhat/lightblue/client/http/LightblueHttpClient.java
+++ b/http/src/main/java/com/redhat/lightblue/client/http/LightblueHttpClient.java
@@ -135,19 +135,21 @@ public class LightblueHttpClient implements LightblueClient {
 
     @Override
     public <T> T data(LightblueRequest lightblueRequest, Class<T> type) throws IOException {
+        LightblueResponse response = null;
         try{
-            LightblueResponse response = data(lightblueRequest);
+            response = data(lightblueRequest);
 
             JsonNode objectNode = response.getJson().path("processed");
 
-            try {
-                return mapper.readValue(objectNode.traverse(), type);
-            } catch (JsonMappingException e) {
-                throw new RuntimeException("Error parsing lightblue response: " + response.getJson().toString(), e);
-            }
+            return mapper.readValue(objectNode.traverse(), type);
         }
-        catch(RuntimeException e){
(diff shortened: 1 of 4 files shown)


dcrissman:
Adding a custom client exception so that client code can distinguish it from other exceptions.

http/src/main/java/com/redhat/lightblue/client/http/LightblueHttpClient.java
```diff
@@ -29,161 +29,167 @@
...
+    @Override
+    public <T> T data(LightblueRequest lightblueRequest, Class<T> type) throws IOException {
+        try{
+            LightblueResponse response = data(lightblueRequest);
+
+            JsonNode objectNode = response.getJson().path("processed");
+
+            try {
+                return mapper.readValue(objectNode.traverse(), type);
+            } catch (JsonMappingException e) {
```
jewzaam: Could this be caught and handled in the outer try/catch?  There's overhead to creating an exception.  In this case, we know the exception is caught and wrapped in yet another exception.

dcrissman: Yes, the inner exception would be wrapped in the outer. I did this because in the case of JsonMappingException, both pieces of information are useful for debugging purposes. However, if the a RuntimeException were to happen from anywhere else in the "data" method, only the outer exception is relevant. 

If you are asking why I didn't just catch JsonMappingException in the outer block, it is primarily because I was trying to prevent duplicate logic. In that I would need to print "Error sending lightblue request:" in both catch blocks.

jewzaam: You don't need to duplicate logic.  How about something like this?  No extra exception is thrown, no duplication of logic, and you get all the data.

```java
catch(RuntimeException | JsonMappingException e) {
    StringBuilder buff = new StringBuilder();
    if (e instanceof JsonMappingException) {
        buff.append("Error parsing lightblue response: " + response.getJson().toString() + "\n");
    }
    buff.append("Error sending lightblue request: " + lightblueRequest.getBody());
    throw new LightblueHttpClientException(buff.toString(), e);
}
```

dcrissman: change submitted.

PR #58, opened 2015-03-06
key to PathValuePair was not getting quotation marks

dcrissman:
SetUpdate $set was also missing quotation marks.


dcrissman:
Fixed how client is handling null RValues

PR #63, opened 2015-03-11
Fixes #57: added impl of client that uses hystrix commands

### LINKED ISSUE #57: Optionally publish hystrix metrics
As more applications start using lightblue and the client, I could easily see a demand for having the option to public hystrix metrics. I know this is something that has already been mentioned as a nice-to-have for the migrator.

jewzaam:
You provide another lightblue client as an argument along with hystrix command and group keys.

Requires merging PR https://github.com/lightblue-platform/lightblue-core/pull/341 first.

Note that this doesn't build right now.  I get an error building I can't figure out yet.

### DIFF
--- a/http/pom.xml
+++ b/http/pom.xml
@@ -5,6 +5,7 @@
     <artifactId>lightblue-client-pom</artifactId>
     <version>1.2.0-SNAPSHOT</version>
   </parent>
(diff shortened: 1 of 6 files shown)


jewzaam:
Ready for review.

PR #64, opened 2015-03-11
add helper methods

dcrissman:


core/src/main/java/com/redhat/lightblue/client/response/LightblueResponse.java
```diff
@@ -40,13 +42,56 @@ public void setJson(JsonNode json) {
...
+    private int parseInt(String fieldName) {
+        JsonNode field = json.findValue(fieldName);
+        if (field == null || field.isNull()) {
+            return 0;
+        }
+        return field.asInt();
+    }
+
+    public <T> T parseProcessed(final Class<T> type) throws LightblueResponseParseException {
+        return parseProcessed(new ObjectMapper(), type);
```
jewzaam: Shouldn't a better configured ObjectMapper be created for parsing JSON?  For example, see https://github.com/lightblue-platform/lightblue-client/blob/85c392811d6b73b250c99438771c99ea97dea5a6/http/src/main/java/com/redhat/lightblue/client/http/LightblueHttpClient.java#L44-L46

dcrissman:
@jewzaam I moved the DEFAULT_MAPPER to the LightblueResponse and then shared it with both.

jewzaam:
@dcrissman constructor of LightblueResponse is creating a new ObjectMapper still.

jewzaam:
Can address ObjectMapper later with https://github.com/lightblue-platform/lightblue-client/issues/65

http/src/test/java/com/redhat/lightblue/client/http/LightblueHttpClientTest.java
```diff
@@ -85,7 +40,7 @@ public void testPojoMappingWithParsingError() throws IOException {
         findRequest.where(withValue("foo = bar"));
         findRequest.select(includeField("_id"));
 
-        String response = "{\"context\": \"rest/FindCommand/esbMessage\", \"errorCode\": \"rest-crud:RestFindError\", \"msg\": \"java.lang.IllegalArgumentException: Cannot call method public static com.redhat.lightblue.crud.FindRequest com.redhat.lightblue.crud.FindRequest.fromJson(com.fasterxml.jackson.databind.node.ObjectNode)\", \"objectType\": \"error\"}";
+        String response = "{\"processed\":\"<p>This is not json</p>\"}";
```
jewzaam: Just a note, but this line is actually valid json.  The intent of the test is to have json that violates the json standard.

dcrissman: I am aware it is technically correct json, but it is not correct for the way we are parsing it.

PR #66, opened 2015-03-12
throw an exception if errors are returned from lb

dcrissman:
This will create a uniformed way for client application code to deal with errors when they are returned from lb.

### DIFF
--- a/core/src/test/java/com/redhat/lightblue/client/response/TestLightblueResponse.java
+++ b/core/src/test/java/com/redhat/lightblue/client/response/TestLightblueResponse.java
@@ -30,7 +30,7 @@ public class TestLightblueResponse {
         new LightblueResponse("bad json");
     }
 
-    @Test(expected = RuntimeException.class)
+    @Test(expected = NullPointerException.class)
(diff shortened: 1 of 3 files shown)


core/src/test/java/com/redhat/lightblue/client/response/TestLightblueResponse.java
```diff
@@ -25,6 +25,17 @@ public void setUp() throws Exception {
         testResponse = new LightblueResponse(initialResponseText);
     }
 
+    @Test(expected = RuntimeException.class)
```
jewzaam: Should tests for specific error expected: LightblueErrorResponseException

jewzaam: I'm assuming it would be this exception, guess you'll let me know if I'm wrong..

dcrissman: No, it should be a RuntimeException. This test is for the constructor which can throw a RuntimeException if an IOException occurs.

Issue #65, opened 2015-03-12
Cleanup new ObjectMapper


jewzaam:
A static DEFAULT_MAPPER was on LightblueHttpClient until https://github.com/lightblue-platform/lightblue-client/pull/64 which moved it to LightblueResponse.  There is also one on a JSON class.  Given DEFAULT_MAPPER sets up some things like date format and it can be reused it would be good to use one instance by default.

dcrissman:
Fixed in #64

jewzaam:
@dcrissman one instance is fixed in that PR, not all:
* https://github.com/lightblue-platform/lightblue-client/blob/1c673a4047dde8d080482301d7d0c9700cc0619f/core/src/main/java/com/redhat/lightblue/client/util/JSON.java#L19
* https://github.com/lightblue-platform/lightblue-client/blob/b64d42c440c771ae2c4217cb68a546cf088ee201/core/src/main/java/com/redhat/lightblue/client/json/LightBlueJsonHelperImpl.java#L15

And maybe this one, but it's a unit test:
* https://github.com/lightblue-platform/lightblue-client/blob/cd2d528dcf9295eac350c0286563569c3ffe09f6/core/src/test/java/com/redhat/lightblue/client/response/TestLightblueResponse.java#L41

dcrissman:
It would make sense to me to move the DEFAULT_MAPPER to JSON.java so that it can be reused.

@jewzaam Are you thinking you'd want the ability to set/change the DEFAULT_MAPPER that the client uses?

jewzaam:
So that for a particular use of the client it could be globally overridden instead of passed around?  May be utility in that, but each has its problems.

dcrissman:
I am thinking both. Set it once and forget (or don't and use DEFAULT_MAPPER) if you want, but LightblueResponse and LightblueHttpClient will continue to have constructors that take an instance of ObjectMapper in case there is need for a one-off that way.

PR #70, opened 2015-03-18
Fixes #28 & Fixes #40: No results issue

### LINKED ISSUE #28: Don't throw exception when no results are returned
No results matching a query provided is not an exceptional situation. Return empty list instead.

### LINKED ISSUE #40: Jackson exception should not be thrown when non JSON is returned.
The lightblue endpoint I was using was down for a bit (due to OpenShift maintenance).  I received an HTML based response, however the stacktrace showed a Jackson parse exception.  Even though this made sense after the fact I feel that it could be handled more gracefully.  Below is an example stacktrace received:

```
com.fasterxml.jackson.core.JsonParseException: Unexpected character ('<' (code 60)): expected a valid value (number, String, array, object, 'true', 'false' or 'null')
 at [Source: java.io.StringReader@2504577; line: 1, column: 2]
	com.fasterxml.jackson.core.JsonParser._constructError(JsonParser.java:1369)
	com.fasterxml.jackson.core.base.ParserMinimalBase._reportError(ParserMinimalBase.java:532)
	com.fasterxml.jackson.core.base.ParserMinimalBase._reportUnexpectedChar(ParserMinimalBase.java:453)
	com.fasterxml.jackson.core.json.ReaderBasedJsonParser._handleUnexpectedValue(ReaderBasedJsonParser.java:1386)
	com.fasterxml.jackson.core.json.ReaderBasedJsonParser.nextToken(ReaderBasedJsonParser.java:669)
	com.fasterxml.jackson.databind.ObjectMapper._initForReading(ObjectMapper.java:2926)
	com.fasterxml.jackson.databind.ObjectMapper._readMapAndClose(ObjectMapper.java:2873)
	com.fasterxml.jackson.databind.ObjectMapper.readTree(ObjectMapper.java:1750)
	com.redhat.lightblue.client.response.LightblueResponse.<init>(LightblueResponse.java:21)
	com.redhat.lightblue.client.http.LightblueHttpClient.callService(LightblueHttpClient.java:119)
	com.redhat.lightblue.client.http.LightblueHttpClient.data(LightblueHttpClient.java:96)
	com.redhat.lightblue.client.http.LightblueHttpClient.data(LightblueHttpClient.java:100)
```

luan-cestari:
Cleaned up code (imports and etc) and created some new tests 

Related to: https://github.com/lightblue-platform/lightblue-client/issues/28

### DIFF
--- a/core/src/main/java/com/redhat/lightblue/client/response/LightblueResponse.java
+++ b/core/src/main/java/com/redhat/lightblue/client/response/LightblueResponse.java
@@ -26,6 +26,8 @@ public class LightblueResponse {
     private JsonNode json;
     private final ObjectMapper mapper;
 
+    // TODO Maybe add a field for some metadata such as HTTP headers and HTTP status code
+
     public LightblueResponse(ObjectMapper mapper) {
         if (mapper == null) {
             throw new NullPointerException("ObjectMapper instance cannot be null");
@@ -69,12 +71,18 @@ public class LightblueResponse {
 
     public boolean hasError() {
         JsonNode objectTypeNode = json.get("status");
-        if (objectTypeNode == null) {
-            return false;
+
+        JsonNode errorCode = json.path("errorCode");
+        if (errorCode != null && !errorCode.isNull() && !errorCode.isMissingNode()){
+            return true;
         }
 
-        return objectTypeNode.textValue().equalsIgnoreCase("error")
-                || objectTypeNode.textValue().equalsIgnoreCase("partial");
+        // it should be true for data from CRUD interface and false from METADATA. Maybe METADATA could also have an status field for every request
+        if((objectTypeNode != null)&&(objectTypeNode.textValue().equalsIgnoreCase("error") || objectTypeNode.textValue().equalsIgnoreCase("partial"))) {
+            return true;
+        }
+
+        return false;
     }
 
     public int parseModifiedCount() {
@@ -96,15 +104,10 @@ public class LightblueResponse {
     @SuppressWarnings("unchecked")
     public <T> T parseProcessed(final Class<T> type)
             throws LightblueResponseParseException {
-        if (hasError()) {
-            throw new LightblueErrorResponseException("Error returned in response: " + getText());
+        if (hasError()){
+            throw new LightblueErrorResponseException("Error parsing lightblue response: Error returned. JSON returned:" + json.toString() + "\n");
         }
-
         try {
-            JsonNode errorCode = json.path("errorCode");
-            if (errorCode != null && !errorCode.isNull() && !errorCode.isMissingNode()){
-                throw new LightblueResponseParseException("Error parsing lightblue response: Error returned. JSON returned:" + json.toString() + "\n");
-            }
 
             JsonNode processedNode = json.path("processed");
 
@@ -121,7 +124,7 @@ public class LightblueResponse {
 
             return mapper.readValue(processedNode.traverse(), type);
         } catch (RuntimeException | IOException e) {
-            throw new LightblueResponseParseException("Error parsing lightblue response: " + getText() + "\n", e);
+            throw new LightblueResponseParseException("Error parsing lightblue response: " + json.toString() + "\n", e);
         }
     }
 
--- a/core/src/main/java/com/redhat/lightblue/client/response/LightblueResponse.java
+++ b/core/src/main/java/com/redhat/lightblue/client/response/LightblueResponse.java
@@ -1,13 +1,13 @@
 package com.redhat.lightblue.client.response;
 
-import java.io.IOException;
-import java.lang.reflect.Array;
-
 import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.redhat.lightblue.client.util.ClientConstants;
+
+import java.io.IOException;
+import java.lang.reflect.Array;
 
 public class LightblueResponse {
 
--- a/core/src/main/java/com/redhat/lightblue/client/response/LightblueResponse.java
+++ b/core/src/main/java/com/redhat/lightblue/client/response/LightblueResponse.java
@@ -96,6 +96,11 @@ public class LightblueResponse {
     public <T> T parseProcessed(final Class<T> type)
             throws LightblueResponseParseException {
         try {
+            JsonNode errorCode = json.path("errorCode");
+            if (errorCode != null && !errorCode.isNull() && !errorCode.isMissingNode()){
+                throw new LightblueResponseParseException("Error parsing lightblue response: Error returned. JSON returned:" + json.toString() + "\n");
+            }
+
             JsonNode processedNode = json.path("processed");
 
             //if null or an empty array
--- a/core/src/main/java/com/redhat/lightblue/client/response/LightblueResponse.java
+++ b/core/src/main/java/com/redhat/lightblue/client/response/LightblueResponse.java
@@ -1,26 +1,19 @@
 package com.redhat.lightblue.client.response;
 
 import com.fasterxml.jackson.databind.DeserializationFeature;
+
+import java.io.IOException;
+import java.lang.reflect.Array;
+
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ArrayNode;
-import com.redhat.lightblue.client.util.ClientConstants;
+import com.redhat.lightblue.client.util.JSON;
 
 import java.io.IOException;
 import java.lang.reflect.Array;
 
 public class LightblueResponse {
-
-    /**
-     * It is safe and encouraged to share the same mapper among threads. It is
-     * thread safe. So, this default instance is static.
-     *
-     * @see <a href="http://stackoverflow.com/a/3909846">The developer of the
-     * Jackson library's own quote.</a>
-     */
-    public static final ObjectMapper DEFAULT_MAPPER = new ObjectMapper()
-            .setDateFormat(ClientConstants.getDateFormat())
-            .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
 
     private String text;
     private JsonNode json;
@@ -36,11 +29,11 @@ public class LightblueResponse {
     }
 
     public LightblueResponse() {
-        this(DEFAULT_MAPPER);
+        this(JSON.getDefaultObjectMapper());
     }
 
     public LightblueResponse(String responseText) {
-        this(responseText, DEFAULT_MAPPER);
+        this(responseText, JSON.getDefaultObjectMapper());
     }
 
     public LightblueResponse(String responseText, ObjectMapper mapper) {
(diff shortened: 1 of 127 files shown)


http/src/main/java/com/redhat/lightblue/client/http/LightblueHttpClient.java
```diff
@@ -119,21 +127,30 @@ public LightblueResponse metadata(LightblueRequest lightblueRequest) {
...
         try {
-            LightblueResponse response = data(lightblueRequest);
-            return response.parseProcessed(type);
-        } catch (RuntimeException | LightblueResponseParseException e) {
-            throw new LightblueHttpClientException("Error sending lightblue request: " + lightblueRequest.getBody(), e);
+            return mapper.readValue(responseNode.traverse(), type);
+        } catch (JsonMappingException e) {
+            LOGGER.error("Error parsing lightblue response: " + responseJson.toString(), e);
+            throw new RuntimeException("Error parsing lightblue response: " + responseJson.toString());
         }
```
jewzaam: This change in data(LightblueRequest, Class<T>) reverts changes to use new LightblueResponse#parseProceseed method.  Why?

http/src/main/java/com/redhat/lightblue/client/http/LightblueHttpClient.java
```diff
@@ -119,21 +127,30 @@ public LightblueResponse metadata(LightblueRequest lightblueRequest) {
...
     public LightblueResponse data(LightblueRequest lightblueRequest) {
         LOGGER.debug("Calling data service with lightblueRequest: " + lightblueRequest.toString());
-        try {
-            return callService(new LightblueHttpDataRequest(lightblueRequest)
-                    .getRestRequest(configuration.getDataServiceURI()));
-        } catch (RuntimeException e) {
-            throw new LightblueHttpClientException("Error sending lightblue request: " + lightblueRequest.getBody(), e);
-        }
+        return callService(new LightblueHttpDataRequest(lightblueRequest)
+                .getRestRequest(configuration.getDataServiceURI()));
```
jewzaam: Method data(LightblueRequest) no longer wraps runtime exception.  I can see why that, at a glance, would seem bad.  Just want to be sure this change doesn't have unexpected side effects.  the difference is the request JSON won't be part of the stacktrace anymore.  I have mixed feelings about this.  1) it would be easier to debug yet 2) it means possible sensitive data is logged on the client.

http/src/test/java/com/redhat/lightblue/client/http/LightblueHttpClientTest.java
```diff
@@ -22,29 +26,93 @@ public void testPojoMapping() throws IOException {
...
+
+        client.setLightblueResponse("{\"status\":\"COMPLETE\",\"modifiedCount\":0,\"matchCount\":0,\"processed\":[]}");
+
+        SimpleModelObject[] results = client.data(findRequest, SimpleModelObject[].class);
+
+        Assert.assertEquals(0, results.length);
+    }
+
+    @Test
+    public void testPojoMappingWithEmptyProcessedResults_ForSingleResult() throws IOException {
```
jewzaam: testPojoMappingWithEmptyProcessedResults_ForSingleResult name is misleading, maybe call it testPojoMappingWithEmptyProcessedResults_ForEmptyJsonNode ?  I see in the test it's verifying the client gracefully handles parsing an empty array to a single null result.

http/src/main/java/com/redhat/lightblue/client/http/LightblueHttpClient.java
luan-cestari: Well,  I think I can change it back how it was and re-throw  the exception in data(LightblueRequest lightblueRequest, Class<T> type)  method, what do you think?

http/src/main/java/com/redhat/lightblue/client/http/LightblueHttpClient.java
luan-cestari: I will revert/fix this

http/src/test/java/com/redhat/lightblue/client/http/LightblueHttpClientTest.java
luan-cestari: I will update with your suggestion

luan-cestari:
Updated the PR. Now almost an LightblueHttpClientException instance is used to wrap the exceptions we caught (only IOException instances are ignored as it is part of the method signature). Added also a line commented why some places the try/catch block was removed as it turn to be redundant for the current implementation

luan-cestari:
Noticed that many classes is missing Copyright header, may I create a PR for this?

http/src/main/java/com/redhat/lightblue/client/http/LightblueHttpClient.java
```diff
@@ -129,11 +131,25 @@ public LightblueResponse data(LightblueRequest lightblueRequest) {
...
         try {
-            LightblueResponse response = data(lightblueRequest);
-            return response.parseProcessed(type);
-        } catch (RuntimeException | LightblueResponseParseException e) {
-            throw new LightblueHttpClientException("Error sending lightblue request: " + lightblueRequest.getBody(), e);
+            return mapper.readValue(responseNode.traverse(), type);
+        } catch (JsonParseException | JsonMappingException e) {
+            LOGGER.error("Error parsing lightblue response: " + responseJson.toString(), e);
+            throw new LightblueHttpClientException("Error parsing lightblue response: " + responseJson.toString());
         }
```
dcrissman: It looks like you are reverting changes made in #64

http/src/main/java/com/redhat/lightblue/client/http/LightblueHttpClient.java
```diff
@@ -157,12 +173,12 @@ protected LightblueResponse callService(HttpRequestBase httpOperation) {
                     HttpEntity entity = httpResponse.getEntity();
                     jsonOut = EntityUtils.toString(entity);
                     LOGGER.debug("Response received from service" + jsonOut);
-                    return new LightblueResponse(jsonOut, mapper);
+                    return new LightblueResponse(jsonOut);
                 }
```
dcrissman: What if a consumer of LightblueHttpClient wants to set a ObjectMapper, how would the LightblueResponse know about it?

http/src/main/java/com/redhat/lightblue/client/http/LightblueHttpClient.java
```diff
@@ -157,12 +173,12 @@ protected LightblueResponse callService(HttpRequestBase httpOperation) {
...
             }
         } catch (IOException e) {
             LOGGER.error("There was a problem calling the lightblue service", e);
-            return new LightblueResponse("{\"error\":\"There was a problem calling the lightblue service\"}", mapper);
+            return new LightblueResponse("{\"error\":\"There was a problem calling the lightblue service\"}");
         }
```
dcrissman: Same question as above.

http/src/main/java/com/redhat/lightblue/client/http/LightblueHttpClient.java
```diff
@@ -129,11 +131,25 @@ public LightblueResponse data(LightblueRequest lightblueRequest) {
...
+        JsonNode objectType = responseJson.get("objectType");
+        if (objectType != null && "error".equalsIgnoreCase(objectType.textValue())) {
+            throw new LightblueHttpClientException("Error in lightblue response: " + responseJson.toString());
+        }
+
+        JsonNode responseNode = responseJson.path("processed");
+        if (responseNode instanceof MissingNode || (responseNode instanceof ArrayNode && responseNode.size() == 0 && !type.isArray())) {
+            return null;
+        }
+
```
dcrissman: Isn't this being handled in LightblueResponse#parseProcessed?

dcrissman:
@luan-cestari I think I see the redundancy you referenced, I submitted #72 to fix. Good catch, thanks.

luan-cestari:
@dcrissman thank for reviewing and nice comments. I just pushed an update. ALso, What do you think about the Copyright thing?

dcrissman:
@luan-cestari I agree about the copywrite?

http/src/main/java/com/redhat/lightblue/client/http/LightblueHttpClient.java
jewzaam: Was the change you made to fix the issue you're trying to fix?  I don't know what is changing in this PR to solve that issue, though I need to do another pass through the changes sometime.

http/src/main/java/com/redhat/lightblue/client/http/LightblueHttpClient.java
```diff
@@ -55,18 +70,15 @@ public LightblueHttpClient(String configFilePath) {
      * This constructor will use a copy of specified configuration object.
      */
     public LightblueHttpClient(LightblueClientConfiguration configuration) {
-        this(configuration, LightblueResponse.DEFAULT_MAPPER);
+        this(configuration, DEFAULT_MAPPER);
     }
```
dcrissman: In #64 I moved this constant to LightblueResponse thinking that we could have a single default mapper defined and use it rather than individual classes defining their own. Maybe a more centralized location would be better? Thoughts?

dcrissman: This is discussed further in #65

luan-cestari: I will revert to previous way, so the PR for the #65 will fix this, ok?

luan-cestari:
Updated the PR

dcrissman:
@luan-cestari Before I merge this, I am fuzzy on what specifically this PR is intended to do. I see a lot of clean up, but would you mind updating the description to include a summary?

dcrissman:
@luan-cestari Could you also merge master in?

luan-cestari:
@dcrissman Updated the issue description and merged with master.

core/src/main/java/com/redhat/lightblue/client/response/LightblueResponse.java
```diff
@@ -101,6 +101,11 @@ private int parseInt(String fieldName) {
         }
 
         try {
+            JsonNode errorCode = json.path("errorCode");
+            if (errorCode != null && !errorCode.isNull() && !errorCode.isMissingNode()){
+                throw new LightblueResponseParseException("Error parsing lightblue response: Error returned. JSON returned:" + json.toString() + "\n");
+            }
+
```
dcrissman: How is this different, or could it be included in hasError?

luan-cestari: I got some NPE in the tests and the exception was pretty similar something I already saw on ligthblue itself, but I confess I didn't look the implementation of has Error , it might be better if there is something missing to put on that method. I will check

luan-cestari: I will try to make a fix o the hasError method, which would include to check errorCode in the json response and I also thought it would be nice to check the HTTP status codes (mainly because if we got the deployment not running and the application server or even the load balancer could return an error and that would not be noticed but that method right now )

luan-cestari:
Updated the PR so hasError could be used stead leaving an error validation somewhere else

alechenninger:
Is there a code style / import order guide for lightblue projects? (re: code cleanup in this PR)

jewzaam:
@dcrissman is this ready to merge?  I see you were reviewing..

@alechenninger http://dev.docs.lightblue.io/formatting/README.html is as close as we have

dcrissman:
My vote is to close this PR.

jewzaam:
I took a look through it and I *think* there are some small changes we might look at cherry picking.  But this PR needs to be closed, the changes if we want them should be implemented in a new PR.

LightblueHttpDataRequest
LightblueHttpMetadataRequest
LightblueHttpClientTest
LightblueResponse
- hasError

dcrissman:
So  much has changed in the client since this PR was created, I don't think there is anything of value here any more. I am going to close it out.

PR #71, opened 2015-03-19
Fixed ValueQuery expressions to allow values with spaces

paterczm:
Also added some expression validation logic.

### DIFF
--- a/core/src/main/java/com/redhat/lightblue/client/expression/query/ValueQuery.java
+++ b/core/src/main/java/com/redhat/lightblue/client/expression/query/ValueQuery.java
@@ -1,20 +1,45 @@
 package com.redhat.lightblue.client.expression.query;
 
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
 import com.redhat.lightblue.client.enums.ExpressionOperation;
 
 /**
  * Created by bmiller on 10/10/14.
+ *
+ * @author bmiller
+ * @author mpatercz
  */
 public class ValueQuery implements Query {
     private String field;
     private String rValue;
     private String operator;
 
+    public static final Pattern expressionPattern = Pattern.compile("(\\w+)\\s*(\\S+)\\s*(.+)$");
+
+    // binary_comparison_operator := "=" | "!=" | "<" | ">" | "<=" | ">=" |
+    //        "$eq" | "$neq" | "$lt" | "$gt" | "$lte" | "$gte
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/redhat/lightblue/client/expression/query/ValueQuery.java
```diff
@@ -1,20 +1,45 @@
...
     private String rValue;
     private String operator;
 
+    public static final Pattern expressionPattern = Pattern.compile("(\\w+)\\s*(\\S+)\\s*(.+)$");
+
+    // binary_comparison_operator := "=" | "!=" | "<" | ">" | "<=" | ">=" |
+    //        "$eq" | "$neq" | "$lt" | "$gt" | "$lte" | "$gte
+    public static final String[] allowedOperators = new String[]{"=", "!=", "<", ">", "<=", ">=", "$eq", "$neq", "$lt", "$lte", "$gte" };
+    public static final Set<String> allowedOperatorsSet = new HashSet<String>(Arrays.asList(allowedOperators));
+
```
dcrissman: Why both the array and the set? Could you merge these two lines?

core/src/test/java/com/redhat/lightblue/client/expression/query/ValueQueryTest.java
```diff
@@ -33,4 +35,25 @@ public void testToStringCallsToJson() throws JSONException {
...
+    @Test(expected=IllegalArgumentException.class)
+    public void testInvalidExpression() {
+        new ValueQuery("fie ld1 = Red Hat Enterprise Linux");
+        Assert.fail();
+    }
+
+    @Test(expected=IllegalArgumentException.class)
+    public void testInvalidOperator() {
+        new ValueQuery("field1 ~= Red Hat Enterprise Linux");
+        Assert.fail();
```
dcrissman: Assert.fail() is not needed as it is already expecting and IllegalArgumentException. If no exception, or the wrong exception is thrown, junit will fail the test.

core/src/test/java/com/redhat/lightblue/client/expression/query/ValueQueryTest.java
```diff
@@ -33,4 +35,25 @@ public void testToStringCallsToJson() throws JSONException {
...
+        String expectedJson = "{\"field\":\"field1\",\"op\":\"=\",\"rvalue\":\"Red Hat Enterprise Linux\"}";
+
+        JSONAssert.assertEquals(expectedJson, expression.toJson(), false);
+    }
+
+    @Test(expected=IllegalArgumentException.class)
+    public void testInvalidExpression() {
+        new ValueQuery("fie ld1 = Red Hat Enterprise Linux");
+        Assert.fail();
+    }
```
dcrissman: I assume this is suppose to fail because there can't be a space in the field name? If so, would you mind dropping a comment above this unit test just to make sure that is clear.

PR #75, opened 2015-03-26
Add license to all java files

dcrissman:


jewzaam:
A while ago we got guidance that putting the copyright in every file was not a requirement.  For that reason I am going to reject this PR.  I'll create an issue with the recommendation.

jewzaam:
See https://github.com/lightblue-platform/lightblue-client/issues/83

PR #77, opened 2015-03-27
fixes #65 - standardizes the default ObjectMapper

### LINKED ISSUE #65: Cleanup new ObjectMapper

dcrissman:
Centralize on a standard default ObjectMapper and allow that to be changed if desired by the client.


jewzaam:
@dcrissman coveralls blew up, maybe rerun the build?  reviewed code and looks good to me

jewzaam:
It fails because of decreasing code coverage.  Going to merge anyway.

PR #79, opened 2015-03-30
fix #78 - add nary operators for values

### LINKED ISSUE #78: Support NaryExpressionOperation operators in ValueQuery
There does not appear to be a way to perform an $in or $nin through any of the Query implementations. It seems to me that ValueQuery would make the most sense.

dcrissman:


dcrissman:
@jewzaam Ready for review

PR #86, opened 2015-04-06
Use enums over string constants

dcrissman:
This touches a bunch of files, but it really only is swapping out string constants for enums for data and metadata operations. The real value in this change is that other implementations of the various operations could be created that reuse the enum values rather the more limiting instanceof checks.

This is a non-breaking change that is part of a larger change I am working on for https://github.com/lightblue-platform/lightblue-migrator/issues/142. For simplicity I have decided to break it up into individual PRs.


dcrissman:
I am debating this change, lets hold off on it for now.

PR #87, opened 2015-04-06
Cleanup request structure

dcrissman:
This fixes an issue where you could call the setBody method, but it would effectively be ignored because getBody was overridden. To me, this is unexpected behaviour and a bug. This also adds a constructor to AbstractLightblueRequest that allows the entityName and entityVersion to be passed up as opposed to implementations having to fend for themselves and creating duplicated logic.


dcrissman:
I know coveralls fails this PR, but really only moving methods around. No net new functionality. 

Ready for review.

paterczm:
Don't all requests have a body? If so, why not keep that field in AbstractLightblueRequest and not override it in concrete implementations?

dcrissman:
They do all have body, the issue is that for AbstractLightblueDataRequest they all override getBody, leaving setBody stranded as a do nothing method. I think that is confusing.

PR #92, opened 2015-04-13
Some clean up and additional factory methods

dcrissman:


dcrissman:
Ready for review, even though coveralls is complaining.

PR #93, opened 2015-04-13
create client test module and move data stubs to it

dcrissman:
This will be useful in other modules testing with the lightblue-client that want to use static json.


dcrissman:
I know coveralls is failing the PR, but that is because I moved some test classes out of the /src/test/java package into it's own module. I don't think it make sense to add unit tests for these test implementations.

paterczm:
-1.29% is not much of a difference.

PR #96, opened 2015-04-23
add substitution ability for property files

dcrissman:
Adding substitution for in-memory lightblue testing.


dcrissman:
We already do something similar to this for json files, so this has some precedence.

https://github.com/lightblue-platform/lightblue-core/blob/master/util/src/main/java/com/redhat/lightblue/util/JsonUtils.java#L65

PR #102, opened 2015-04-30
Inject an http client interface (fixes #42)

### LINKED ISSUE #42: Inject an HttpClient instead of new-ing it up internally inside LightblueHttpClient

alechenninger:
Still more to do, but this is a basic idea of what I was campaigning for in #42.

I'm going to be critical of the existing implementation so, forewarning and I mean nothing personal. Open to critique of all kinds on this refactoring as well.

There were problems, though, and there still are, but here's a summary:

- LightblueRequest is a better interface now. Before it was muddied between details that were specific to Lightblue and specific to HTTP. Ultimately, a lightblue request is like any other HTTP request: it has a URI, it has a body (optionally), and it has an HTTP method. That's it. This clarifies that intent and pushes details that are only relevant to implementations, down to implementations. It also gets the missing piece of the http request: the method. Before you had URI and body, but not method, so the method had to be derived from massive select statements elsewhere in the code, orthogonal to where the rest of information about the request lives. Now, the configuration of a lightblue request is all a part of each request's implementation, and nothing else.
- The client itself accepts an actual httpclient as a constructor argument (classic dependency injection). This means this client is now testable, and also very simple. The specifics of an http client can be tested in isolation, and now the implementations can grow.
- For this, there is a very very simple http client that satisfies what we needs, but can grow if we need it to in the future. It can be implemented by anything that can make http requests very easily.
- And so, the apache http client usage is moved out to its own class which implements this interface. This means we can retain existing compatibility with client behavior (does not need to be closed), but allow the client to use other, more efficient implementations which may requiring lifecycle management if a user chooses (and hopefully this can be default in the future).

Still need to clean up dead code, unit test, etc...


core/src/test/java/com/redhat/lightblue/client/response/TestLightblueResponse.java
```diff
@@ -16,25 +16,25 @@
 
 public class TestLightblueResponse {
```
dcrissman: Nitpick, but lets change this to TestDefaultLightblueResponse just for clarity.

core/src/main/java/com/redhat/lightblue/client/LightblueClient.java
```diff
@@ -4,14 +4,15 @@
...
 public interface LightblueClient {
 
-    public abstract LightblueResponse metadata(AbstractLightblueMetadataRequest lightblueRequest);
+    LightblueResponse metadata(LightblueRequest lightblueRequest);
 
-    public abstract LightblueResponse data(AbstractLightblueDataRequest lightblueRequest);
+    LightblueResponse data(LightblueRequest lightblueRequest);
 
-    public abstract <T> T data(AbstractLightblueDataRequest lightblueRequest, Class<T> type) throws IOException;
+    <T> T data(AbstractLightblueDataRequest lightblueRequest, Class<T> type) throws IOException;
```
dcrissman: I see what you are doing and I am fine with switching this, but I would like to see LightblueHttpDataRequest and LightblueHttpMetadataRequest throw an exception instead of returning a null value.

dcrissman:
+1 I like what you are doing.

As a suggestion, I would like to see the PATH_PARAM_XXX parameters not stored as on the respective parent classes. I think only the implementations need to know what their specific value is.

alechenninger:
> As a suggestion, I would like to see the PATH_PARAM_XXX parameters not stored as on the respective parent classes. I think only the implementations need to know what their specific value is.

Yeah definitely, I can't understand why it was done that way. There's still a lot that could be improved, this is mostly a POC at the moment, but still any feedback is valuable. I just haven't looked at a lot yet.

dcrissman:
@alechenninger Would you merge in changes from master?

Are you ready for this to be code reviewed and merged? Or do you want to do more work on it first?

alechenninger:
Need some more time to work on this guy, sorry for the lack of updates. I should be able to finish this up this weekend. I'll @ mention someone when it's ready. Thank you!

alechenninger:
Did some benchmarks against an SSL lightblue... using the java.net implementation saw an average of **70%** performance increase over the original, self-closing Apache client implementation. Average of around 500ms improvement. /cc @paterczm

core/src/main/java/com/redhat/lightblue/client/response/DefaultLightblueResponse.java
```diff
@@ -0,0 +1,127 @@
...
+import java.lang.reflect.Array;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.redhat.lightblue.client.util.JSON;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DefaultLightblueResponse implements LightblueResponse {
```
alechenninger: This is the same as the old `LightblueResponse`, I just pulled out an interface for it and renamed the old implementation to `DefaultLightblueResponse`. This is so the top level interface could be immutable (not have setters).

core/src/main/java/com/redhat/lightblue/client/response/LightblueResponse.java
```diff
@@ -1,121 +1,17 @@
...
-import java.lang.reflect.Array;
-
 import com.fasterxml.jackson.databind.JsonNode;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.node.ArrayNode;
-import com.redhat.lightblue.client.util.JSON;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-public class LightblueResponse {
```
alechenninger: As you can see here is the old one

http/src/main/java/com/redhat/lightblue/client/http/LightblueHttpClient.java
```diff
@@ -18,17 +14,17 @@
...
 import com.redhat.lightblue.client.request.AbstractLightblueDataRequest;
-import com.redhat.lightblue.client.request.AbstractLightblueMetadataRequest;
+import com.redhat.lightblue.client.request.LightblueRequest;
+import com.redhat.lightblue.client.response.DefaultLightblueResponse;
 import com.redhat.lightblue.client.response.LightblueResponse;
 import com.redhat.lightblue.client.response.LightblueResponseParseException;
 import com.redhat.lightblue.client.util.JSON;
 
-public class LightblueHttpClient implements LightblueClient {
+public class LightblueHttpClient implements LightblueClient, Closeable {
```
alechenninger: The internal, Apache http client specific were gutted for a generic `HttpClient` interface that just has an `executeRequest` method. This class also by default uses a java.net implementation of the new `HttpClient` interface which is, in my tests, around 70% faster for SSL requests.

http/src/main/java/com/redhat/lightblue/client/http/LightblueHttpClient.java
```diff
@@ -106,10 +112,13 @@ public LightblueHttpClient(String dataServiceURI, String metadataServiceURI, Boo
      * .client.request.LightblueRequest)
      */
     @Override
-    public LightblueResponse metadata(AbstractLightblueMetadataRequest lightblueRequest) {
+    public LightblueResponse metadata(LightblueRequest lightblueRequest) {
```
alechenninger: I removed the `AbstractLightblue*Request` classes since they were no longer needed. Before they were used to encapsulate the HTTP method, but that's been moved to live with the rest of the request details in `LightblueRequest`. This means you can technically pass a request meant for a data node to `metadata`. If you don't like this we can add a marker interface to distinguish between metadata requests and data requests.

http/src/main/java/com/redhat/lightblue/client/http/request/AbstractLightblueHttpRequest.java
```diff
@@ -1,41 +0,0 @@
...
-
-import java.io.UnsupportedEncodingException;
-
-import org.apache.http.client.methods.HttpDelete;
-import org.apache.http.client.methods.HttpGet;
-import org.apache.http.client.methods.HttpPost;
-import org.apache.http.client.methods.HttpPut;
-import org.apache.http.entity.StringEntity;
-
-public class AbstractLightblueHttpRequest {
```
alechenninger: All of these were Apache specific and no longer needed

http/src/main/java/com/redhat/lightblue/client/http/request/LightblueHttpRequest.java
```diff
@@ -1,9 +0,0 @@
-package com.redhat.lightblue.client.http.request;
-
-import org.apache.http.client.methods.HttpRequestBase;
-
-public interface LightblueHttpRequest {
```
alechenninger: A `LightblueRequest` is and always was an HTTP request, so there was no need for a separate `LightblueHttpRequest` AFAIK. The only thing `LightblueRequest` was missing was the HTTP method, but it had the URI (path params) and request body. I suppose request body could be used to talk to lightblue another way other than HTTP if we really wanted to be that generic but, not sure what URI would be for if not over HTTP, so yeah, that's why I consolidated those interfaces.

http/src/main/java/com/redhat/lightblue/client/http/transport/HttpClient.java
```diff
@@ -0,0 +1,11 @@
+package com.redhat.lightblue.client.http.transport;
+
+import com.redhat.lightblue.client.request.LightblueRequest;
+
+import java.io.Closeable;
+import java.io.IOException;
+
+public interface HttpClient extends Closeable {
+    // TODO: Do we need to able to examine headers or status code of response?
```
alechenninger: Just returned String here since that's all that's technically need to keep 1:1 with existing behavior, but we may want to consider returning an `HttpResponse` object that has status code and other information from the request. This would probably be fine to be rolled into `LightblueResponse`, in which case this method would return that. Something to consider. For now this is simple and it works.

http/src/main/java/com/redhat/lightblue/client/http/transport/SelfClosingApacheHttpClient.java
```diff
@@ -0,0 +1,46 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.security.GeneralSecurityException;
+
+/**
+ * A simple to use though inefficient http client which does not require closing. Useful for simple
+ * applications or one off requests.
+ */
+public class SelfClosingApacheHttpClient implements HttpClient {
```
alechenninger: This should be deprecated or removed I think. Not sure if there's any reason to use this over the java.net implementation.

http/src/test/java/com/redhat/lightblue/client/http/LightblueHttpClientMock.java
```diff
@@ -1,25 +0,0 @@
-package com.redhat.lightblue.client.http;
-
-import org.apache.http.client.methods.HttpRequestBase;
-import com.redhat.lightblue.client.LightblueClientConfiguration;
-
-import com.redhat.lightblue.client.response.LightblueResponse;
-
-public class LightblueHttpClientMock extends LightblueHttpClient {
```
alechenninger: Test subclass no longer needed because you can inject a `mock(HttpClient.class)`

http/src/test/java/com/redhat/lightblue/client/http/request/AbstractLightblueHttpRequestTest.java
```diff
@@ -1,38 +0,0 @@
...
-import java.io.IOException;
-
-import org.apache.commons.io.IOUtils;
-import org.apache.http.client.methods.HttpPost;
-import org.apache.http.client.methods.HttpPut;
-import org.apache.http.client.methods.HttpRequestBase;
-import org.apache.http.util.EntityUtils;
-import org.junit.Assert;
-
-public class AbstractLightblueHttpRequestTest {
```
alechenninger: All of those classes were removed, these are the tests

http/src/test/java/com/redhat/lightblue/client/http/transport/JavaNetHttpClientIntegrationTest.java
```diff
@@ -0,0 +1,67 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+import java.io.IOException;
+
+@RunWith(JUnit4.class)
+public class JavaNetHttpClientIntegrationTest {
+    @Rule
+    public WireMockRule wireMockRule = new WireMockRule();
```
alechenninger: Wiremock [very quickly] starts an http server for the tests. This is to test the real semantics of java.net HttpUrlConnection which I wasn't totally familiar with (how error responses are dealt with in particular).

http/src/test/java/com/redhat/lightblue/client/http/testing/doubles/FakeLightblueRequest.java
```diff
@@ -0,0 +1,31 @@
+package com.redhat.lightblue.client.http.testing.doubles;
+
+import com.redhat.lightblue.client.http.HttpMethod;
+import com.redhat.lightblue.client.request.LightblueRequest;
+
+public class FakeLightblueRequest implements LightblueRequest {
```
paterczm: Why is LightblueRequest implementation called fake?

Those changes move http protocol stuff down from lightblue http client to transport. What is your recommendation for manipulating request headers? Example: I want to set Accept-Encoding header depending on lightblue configuration. This should belong to LightblueHttpClient class, because it's independent on the underlying http client implementation.

paterczm: It's fake because it's  for testing... never mind the first question :/

alechenninger: To the second question, I don't necessarily think it should belong to the lightblue client class. At this point I see the lightblue client class as really just a dumb wrapper around whether or not a request should go to a data node or a metadata node, and turning the String response into a LightblueResponse object. The details of the HTTP request I think should be in the http client (too many "clients"!). I would say since we know Lightblue Rest is configured to compress, then we can allow configuring the clients to accept that encoding as well. The most naive implementation would be a "shouldCompress" flag in the constructor for that `HttpClient` impl, and if this is set, the accept-encoding header is set appropriately, and the http client itself deals with decompressing (it's contract is that it gets to the `LightblueHttpClient` class as a normal uncompressed string). How the http client deals with the encoding would be impl specific.

Do you see issues with that approach?

paterczm: The issues I see:
* As you said, too many clients: lightblue client, specific http client implementation wrapper and specific http client implementation. First 2 layers are thin wrappers and carry little substance.
* Implementing header handling for each http client implementation feels redundant, though perhaps it's unavoidable if multiple http client implementations are to be supported.
* Actually, I don't see much benefit of being able to switch http client implementations. I would decide on one and stick to it (it could still be changed if need be, just not so conveniently).

So far I implemented compression functionality as you suggested. This is not a big deal, I just have an impression that you are over-generalizing things a bit. If someone said they want to be able to switch http clients, then I'd rest my case.

paterczm:
@alechenninger, I'd like to wrap it up, having changes like this unmerged is asking for trouble (already run into conflicts with compression branch). Though I don't see much benefit of providing yet another layer of abstraction to support different http client implementations, I'm not very opposed to the idea either. Those changes also include better http client lifecycle with keep-alive support, which works great . I want it in master. What else needs to be done before we can merge?

alechenninger:
@paterczm Just wanted to add some tests and clean up any concerns anyone had. I'm good with merging if you are. I could rename the "http clients" to "http transports" if you want, or something like that. That's probably a good last thing to do to prevent confusion. What do you think?

I share some of your concerns and want to echo them as a bit of maintenance caution. I'm less after being able to switch out implementations and more just wanted dependency injection in general which as a side effect allows you to switch out "real" implementations also. Being able to do so is mainly important for testability. The java.net client was motivated purely by being able to keep existing ease of use and not break current usage (not requiring to close the client), and also get persistent connections. And now the apache version just exists because it already did and folks may want an environment where they can configure a thread pool or whatever other Apache feature they need (like a proxy, etc.). but now to do that they just have to make sure they close out the client at some point.

Anyhow, I'm giving this backstory to say: I didn't make it generic to get into the business of making generic http client implementations. It's that way to be testable and to provide a minimal alternate implementation for ease of use. If our demands on the http transport start to broaden, the scope of the java.net implementation should be managed as such. Perhaps it would eventually even be deprecated in favor of just configuring Apache and telling people to remember to close it. So I see the java.net as a minimal stop-gap measure when nothing fancier is needed, not our own version of Apache's already full featured client.

Anyhow, let me know if you think we should rename some things for sanity's sake, otherwise merge at will! :)

alechenninger:
Oh, another quick question -- the "self closing Apache client" -- I think we should remove this since the java.net client does that but without the massive performance hit. What do you think?

paterczm:
>  I could rename the "http clients" to "http transports" if you want, or something like that. That's probably a good last thing to do to prevent confusion. What do you think?

Yes, that would help distinguish the layers.

I think I understand why you decided to use java.net client and that testability requires injectability. However, this does not mean we need to support injecting different http client implementations. If we are to continue with java.net and apache, then both wrappers will need unit tests, both should have lzf compression support added (and again, unit tests...). It adds some overhead to maintain a feature we don't need.

I propose to focus on java.net / JavaNetHttpClient and get the apache client wrapper out of the way (annotate is as Beta or remove it completely).

http/src/main/java/com/redhat/lightblue/client/http/transport/JavaNetHttpClient.java
```diff
@@ -0,0 +1,169 @@
...
+    /**
+     * Parses response, whether or not the request was successful, if possible. Reads entire input
+     * stream and closes in order to so the socket knows it is finished and may be put back into a
+     * pool for reuse.
+     */
+    private String response(HttpURLConnection connection) throws IOException {
+        try (InputStream responseStream = connection.getInputStream()) {
+            return readResponseStream(responseStream, connection);
+        } catch (IOException e) {
+            try (InputStream errorResponseStream = connection.getErrorStream()) {
```
paterczm: I think this error should be logged to server.log.

paterczm:
I'm done with lzf compression support for JavaNetHttpClient. Will create a pr once those changes are merged.

alechenninger:
Awesome. Sounds good. Personally I would keep Apache option around but I see your point :). Would you like me to remove _everything_ Apache with this PR? Once Apache's client is not a supported option there is a lot of code that could be removed: almost everything under auth package, most of the code in SslSocketFactories, dependency from pom, maybe something else. Or I could just remove the Apache wrapper I added for now and remove everything else in a separate PR.

http/src/main/java/com/redhat/lightblue/client/http/transport/JavaNetHttpClient.java
alechenninger: Should that be responsibility of application? Or perhaps the `LightblueHttpClient` can examine response object for error codes and throw appropriate exceptions with response body and request info. I think #66 accomplished this (though I feel it's weird that an exception is only thrown when trying a client tries to use the response body, and not before the response is returned from the client).

alechenninger:
Can't remove all of the Apache stuff just yet actually because the proxy servlets rely on it. The only advantage of using Apache there is, aside from it not having the problem of closing sockets prematurely, Apache client supports streaming the response bodies (our HttpTransport iface returns a String), and streaming would be better performance for the servlet to use (won't have to redundantly wait to load the whole response into memory, then stream).

paterczm:
Having 2 http client implementations, each used for something else - it's exactly what I'm trying to avoid. I'm merging this now and will create a separate issue to unify proxy servlet and client.

PR #103, opened 2015-05-01
Setting content-type to 'application/json'

paterczm:
GET requests work, but other methods fail. I see the content-type header being set in firebug, so I assume the proxy is to blame for not passing it down to lightblue.


paterczm:
@alechenninger, fyi

alechenninger:
Yeah hrm I see. Proxy does not copy any headers from original request. Do you think it should? It's probably reasonable just to set content type header.

Would you be able to add a small unit test to verify this header is set on the outgoing request?

alechenninger:
The test setup is somewhat annoying to stub out a working servlet request and response object, so it may be excusable to add it as part of this existing test: `AbstractLightblueProxyServletTest#shouldProxyTheRequestMethodUriAndBody`

PR #118, opened 2015-05-22
Adding lzf compression ability to the client

paterczm:


core/src/main/java/com/redhat/lightblue/client/LightblueClientConfiguration.java
```diff
@@ -6,6 +6,21 @@
...
+        NONE, LZF;
+
+        public static Compression parseCompression(String str) {
+            switch (str) {
+            case "NONE":
+                return NONE;
+            case "LZF":
+                return LZF;
+            default:
+                throw new IllegalArgumentException("Invalid compression " + str + ". Supported values are: NONE, LZF");
```
dcrissman: Should NONE just be the default for backwards compatibility purposes?
Are there any legitimate reasons someone would not want to use compression? If not, then maybe that should just be the default?

paterczm: The protocol ensures backwards compatibility. If the client advertises ability to decompress responses, but lightblue doesn't set content-encoding header to lzf, then the client will treat the response as not compressed.

Lzf is considered to be very fast and I doubt it will slow things down in any scenario (even a very fast network). On slower networks it will offer better response times.

For those reasons, I made compression on by default.

dcrissman: What happens if the client advertises the ability, but NONE was set on the server? I assume lzf would not be used. Would the client still be able to read the response? Or is it expecting lzf to be returned?

paterczm: I'm not sure if you can configure that server side, but if the server does not flag the response as compressed with a special header, then the client will not try to decompress it. It works the same the other way around - if the client does not set a special header to indicate it's capable to decompress lzf, lightblue will not compress responses.

dcrissman: Fair, merging.

PR #119, opened 2015-05-26
remove stub requests

dcrissman:
With the addition of https://github.com/lightblue-platform/lightblue-client/blob/master/core/src/main/java/com/redhat/lightblue/client/expression/query/LiteralQuery.java, these test classes should no longer be needed.


dcrissman:
Still used in lightblue-client's AbstractLightblueClientCRUDController

PR #132, opened 2015-07-09
Update lightblue-client error handling. It throws now a LightblueExce…

ykoer:
…ption containing the Errors and DataErrors


bserdar:
Few things:
 1) Error.java needs cleanup. No need for logging, threadlocals, static
get() methods, constructors push/pup, and toJson()
 2) Add a LightblueResponse instance to the LightblueException, instead of
just the errors. Otherwise, there is no way to get the response. errorCodes
set can be initialized lazily.
 3) There is no need for JsonObject class. And there is no need for toJson
methods. Errors are already parsed from json.

core/src/main/java/com/redhat/lightblue/client/model/Error.java
```diff
@@ -0,0 +1,76 @@
...
+ * client as an indicator of where the error happened.
+ * 
+ */
+public final class Error extends RuntimeException {
+
+    private static final long serialVersionUID = 1L;
+
+    public static final char DELIMITER = '/';
+
+    private final ArrayDeque<String> context;
```
bserdar: String context should be enough.

core/src/main/java/com/redhat/lightblue/client/model/Error.java
```diff
@@ -0,0 +1,76 @@
...
+
+    public String getErrorCode() {
+        return errorCode;
+    }
+
+    public String getMsg() {
+        return msg;
+    }
+
+    public void pushContext(String context) {
```
bserdar: No need for this once context is string

core/src/main/java/com/redhat/lightblue/client/model/Error.java
```diff
@@ -0,0 +1,76 @@
...
+            }
+            x = ((ObjectNode) node).get("msg");
+            if (x != null) {
+                m = x.asText();
+            }
+
+            Error ret = new Error(e, m);
+
+            x = ((ObjectNode) node).get("context");
+            if (x != null) {
```
bserdar: And you can get rid of this as well. Context is simply a string for the client

core/src/main/java/com/redhat/lightblue/client/response/LightblueException.java
```diff
@@ -0,0 +1,257 @@
...
+    public void setLightblueResponse(DefaultLightblueResponse lightblueResponse) {
+        this.lightblueResponse = lightblueResponse;
+    }
+
+    public boolean exists(String errorCode) {
+        return errorCodes.contains(errorCode);
+    }
+
+    public Set<String> getErrorCodes() {
+        return errorCodes;
```
bserdar: initialize error codes here, instead of in the constructor

ykoer: I would then have to initialize the error codes in "exists" method as well, which I don't like.

bserdar: You could do getErrorCodes().contains(errorCode)

Not a big deal though.

PR #144, opened 2015-08-05
New expression classes

bserdar:
This PR rewrites query, projection, sort, and update classes with a uniform API and JSON back-end instead of a string back-end. It also modifies the existing request class structure to support both the old deprecated queries, and the new ones.


core/src/main/java/com/redhat/lightblue/client/ExpressionPart.java
```diff
@@ -0,0 +1,26 @@
...
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.fasterxml.jackson.databind.node.JsonNodeFactory;
+
+/**
+ * Base class for any expression part that is backed by a Json node
+ */
+public abstract class ExpressionPart implements JsonObj {
+    protected JsonNode node;
```
dcrissman: Should node be final?

bserdar: It should be, but just because of Literal(Object x), it can't be.

bserdar: ...I suppose that can be written differently, allowing it to be final. Let
me try..

core/src/main/java/com/redhat/lightblue/client/Expression.java
```diff
@@ -0,0 +1,52 @@
...
+    protected Expression(boolean arrayNode) {
+        super(arrayNode?JsonNodeFactory.instance.arrayNode():
+              JsonNodeFactory.instance.objectNode());
+    }
+
+    /**
+     * Add field:value
+     */
+    public Expression add(String field,JsonNode value) {
+        ((ObjectNode)node).set(field,value);
```
dcrissman: I would say do an instanceof check and throw a meaningful exception.

bserdar: Rewrote the exception as runtime exception.

core/src/main/java/com/redhat/lightblue/client/Query.java
```diff
@@ -0,0 +1,245 @@
...
+
+    public static final NaryOp in=NaryOp.in;
+    public static final NaryOp nin=NaryOp.nin;
+
+    public static final ArrOp any=ArrOp.any;
+    public static final ArrOp all=ArrOp.all;
+    public static final ArrOp non=ArrOp.none;
+
+    public static final LogOp and=LogOp.and;
+    public static final LogOp or=LogOp.or;
```
dcrissman: You already have these defined as enums. It seems weird to me to then put them into constants.

bserdar: Convenience: allows writing Query.cmp("field",Query.eq,1)

dcrissman: I still think it is unnecessary, but not enough to block the PR.

bserdar: Operators are now inner classes, so it is now much easier to write
Query.eq, instead of Query.BinOp.eq. Also, this way you don't need to know
a BinOp or NaryOp exists, all you need to know is eq. Simply importing
Query will bring in all the names necessary to build all queries.

core/src/main/java/com/redhat/lightblue/client/request/data/LiteralDataRequest.java
```diff
@@ -12,29 +15,46 @@
...
+    public LiteralDataRequest(String entityName, String entityVersion, JsonNode body, HttpMethod httpMethod, String operationalPathParam) {
+        super(entityName, entityVersion);
         this.body = body;
         this.httpMethod = httpMethod;
         this.operationPathParam = operationalPathParam;
     }
 
     @Override
-    public String getBody() {
+    public JsonNode getBodyJson() {
```
dcrissman: This is a potentially breaking change. I have no objections, but if we are going to force this, does it make just as much sense to rip out the other depreciated classes/enums and just rip the bandage off?

bserdar: Why is it a breaking change? getBody() still exists (coming from the abstract base), and returns string.

dcrissman: Fair enough.

paterczm:
@bserdar, @dcrissman 

Are there any usage examples of the new api? Looking at the code, you're replacing 
```
withValue("field = value")
```
with
```
cmp("field", Query.eq, "value")
```

I find the now deprecated parsed expression more readable and concise. I understand the need for json backend, but I would not change the api. It's critical for lightblue adoption.

dcrissman:
@paterczm makes a good point. I agree.

bserdar:
Fair point. I'll rename the new APIs to match the old ones, and add support
for the case you mentioned.

PR #149, opened 2015-08-13
allow for easily parsing just the processed json body

dcrissman:


core/src/main/java/com/redhat/lightblue/client/response/DefaultLightblueResponse.java
```diff
@@ -141,6 +141,10 @@ private int parseInt(String fieldName) {
         return field.asInt();
     }
 
+    public JsonNode parseProcessed() {
+        return json.path("processed");
```
bserdar: I'd rename this method to getProcessed(), no parsing is done here.

Also, it might make more sense to use json.get('processed"), as that will return null if "processed" does not exist.

dcrissman:
@bserdar requested changes made

PR #150, opened 2015-08-19
allow update to accept Date

dcrissman:


core/src/main/java/com/redhat/lightblue/client/Update.java
```diff
@@ -249,6 +251,10 @@ public static Set set(String field,String i) {
     public static Set set(String field,boolean i) {
         return new Set().more(field,Literal.value(i));
     }
+    
+    public static Set set(String field, Date date){
+        return set(field, ClientConstants.getDateFormat().format(date));
```
bserdar: Insted, use set(field,Literal.value(date))

dcrissman:
@bserdar Changes made.

PR #153, opened 2015-08-24
use ThreadLocal instead of clone

dcrissman:


dcrissman:
@jewzaam This PR should fix the issue you mentioned from http://www.javacodegeeks.com/2010/07/java-best-practices-dateformat-in.html

dcrissman:
Moving to in-progress as there seems to be additional dialog.

jewzaam:
As noted in the linked article there is no benefit with this model if there's no thread pool.  Do you know if a pool is used?  There is no cleanup as well.  Is there a way to time out old / unused records in ThreadLocale?  

Guess this may require some more thought.  What about a cache/pool of SDF's?  Something that would grow to a limited size and allow object reuse.  Can't remember what I've used in the past, but there's probably a few good frameworks out there.  This all assumes we have enough volume of events and concern with performance that tackling this now makes sense.

bserdar:
What is wrong with clone() from a static instance? Clone is not an
expensive operation.

dcrissman:
I am not specifically using any kind of thread pool. I would assume camel is using some sort of pool, but I don't know the specifics.

Close?

jewzaam:
If clone is cheap go with clone. Could compare time to create new simple
date format to a clone operation?

bserdar:
So I ran a test. Cloning is about 5 times faster than creating new
SimpleDateFormat. Retrieving from a thread local is about 100 times faster
than clone. Turns out clone is not that cheap.

The ultimate solution would be to clone from a static copy to initialize a
thread local.

dcrissman:
Changes made to use a combination strategy of both ThreadLocal and clone.

jewzaam:
This is setting us up for a "memory leak" if there is no thread pooling or if the JVM is online very long.  There is no destruction of the SDF's that are created.  This implementation assumes a thread pool and no thread deaths.

bserdar:
Thread local should clean up date formats as threads die. It is using a
weak hash map to store per-thread values, so when a thread is discarded,
the date format should lose its reference.

Issue #143, opened 2015-08-03
Duplicate Response Exceptions


dcrissman:
It seems that we now have two exceptions that could occur for a lightblue response from the lightblue-client, they seem to mean similar things: LightblueException & LightblueErrorResponseException. I don't feel that we need both.

LightblueErrorResponseException could probably be eliminated as the DefaultLightblueResponse's constructor now throws LightblueException is an error is on the response.

That said, LightblueException has a laundry list of error codes that I am not sure why they are there. I am concerned that this list will quickly become unmaintained and I question the value.

What uses the constant error codes and are there any good reason why they should not be removed?

bserdar:
The error codes are there because there is not a list of exception codes
publicly available anywhere else. Not all exceptions end up being reported
in the same way. Certain exceptions are handled differently, either by
rethrowing a specific exception, or consumed as part of the service logic.
Those cases need to know error codes for certain errors, such as "duplicate
key".

So, the exception codes should stay, but they could be moved out to a
separate constants class.

The fact that it will be unmaintained is not a big problem. If you need to
check the existence of an error code and it is not there, then you add it,
effectively maintaining it.

dcrissman:
I am not opposed to making the constants available somehow through the client. I would feel better about having the codes in a constants class. 

I am still concerned about a long list of unmaintained codes though. Say we go the other way and a code is removed from some controller, but not removed from the list. That could cause some confusion.

Maybe getting to a bigger issue I have though, being that the error codes themselves are not always clear as to what exactly they mean or under what circumstances they can fire. Having constants is great, but not helpful if you don't know what they mean. This is perhaps outside of scope, but I think we need to make sure codes are standard and consistent across modules. Some documentation (dun-dun-dun) might be helpful as well.

dcrissman:
This was addressed in #189. Closing.

PR #155, opened 2015-09-08
Exception error codes

snehagunta:
Pulled exception error code constants into a separate class #143. Addressed the issue with error constants in Exception class


core/src/main/java/com/redhat/lightblue/client/response/LightBlueResponseErrorCodes.java
```diff
@@ -0,0 +1,191 @@
+package com.redhat.lightblue.client.response;
+
+public class LightBlueResponseErrorCodes {
```
bserdar: Pls make this an interface, and have the exception class implement it for now. There are clients using the constants in the exception class.

PR #156, opened 2015-09-08
Fixes #154 adds bulk support to Java client

### LINKED ISSUE #154: Support Bulk Request
Feature recently added to lightblue-core and lightblue-rest.

BVulaj:


core/src/main/java/com/redhat/lightblue/client/request/BulkLightblueDataRequest.java
```diff
@@ -0,0 +1,61 @@
...
+				op = "delete";
+			} else if (req instanceof DataFindRequest) {
+				op = "find";
+			} else if (req instanceof DataInsertRequest) {
+				op = "insert";
+			} else if (req instanceof DataSaveRequest) {
+				op = "save";
+			} else if (req instanceof DataUpdateRequest) {
+				op = "update";
+			}
```
dcrissman: I would suggest calling getOperationPathParam() instead of this method having to know what all the possible operations are.

dcrissman:
> Don't use JSON.toJsonNode. Instead, use JsonNodeFactory.instance.numberNode() or textNode()/
@bserdar What about using JSON.getDefaultObjectMapper().getNodeFactory()? This would then use the same instance as the configured ObjectMapper. We could create a short cut method on JSON if that would help.

core/src/main/java/com/redhat/lightblue/client/request/BulkLightblueDataRequest.java
BVulaj: That is what I had prior, but @bserdar suggested I change it.  What's the consensus?

bserdar:
What's the point of going through object mapper and two getters instead of
a singleton variable instance?

dcrissman:
@BVulaj We typically use spaces instead of tabs. I see a lot of your changes switch to tabs. Makes it hard to see what changed. I would prefer you switch back to spaces.

core/src/main/java/com/redhat/lightblue/client/request/BulkLightblueDataRequest.java
bserdar: Just because it returns the string you want doesn't make it readable or
maintainable. getOperationPathParam returns a string that is intended to be
a part of a URL. Such implicit dependencies between unrelated parts of code
will cause trouble when things change.

dcrissman:
> What's the point of going through object mapper and two getters instead of a singleton variable instance?

@bserdar I missed it was a singleton instance. So no point then.

BVulaj:
@dcrissman I noticed the some of the changes had spaces, but I wasn't sure if it was a team convention or anything.  I can switch from tabs to spaces.

core/src/main/java/com/redhat/lightblue/client/LightblueClient.java
```diff
@@ -1,18 +1,22 @@
...
-    LightblueResponse metadata(LightblueRequest lightblueRequest);
+	Locking getLocking(String domain);
 
-    LightblueResponse data(LightblueRequest lightblueRequest) throws LightblueException;
+	LightblueResponse metadata(LightblueRequest lightblueRequest);
 
-    <T> T data(AbstractLightblueDataRequest lightblueRequest, Class<T> type) throws LightblueException;
+	LightblueResponse data(LightblueRequest lightblueRequest) throws LightblueException;
+
+	BulkLightblueResponse dataBulk(AbstractBulkLightblueRequest<AbstractLightblueDataRequest> request) throws LightblueException;
```
bserdar: I think data() would be just fine, instead of dataBulk

core/src/main/java/com/redhat/lightblue/client/request/BulkLightblueDataRequest.java
dcrissman: Fair.

What if we made the op the simple name of the class? Then there would be no need to maintain the hard code condition.

core/src/main/java/com/redhat/lightblue/client/LightblueClient.java
BVulaj: @bserdar I changed this since method selection was somewhat ambigious, since AbstractBulkLightblueRequest implements LightblueRequest.  I'll take a look at changing it back to data() and fixing any ambiguities in its use.

bserdar: At least make it bulkData, instead of dataBulk

core/src/main/java/com/redhat/lightblue/client/request/BulkLightblueDataRequest.java
bserdar: They are still semantically two different values.

The way I'd prefer would be to have a getOperation() method in the Request
base class.

dcrissman: That was going to be my next suggestion.

+1 to getOperation()

bserdar: +1 as well, and either constant strings, or an enum for operations.

core/src/main/java/com/redhat/lightblue/client/LightblueClient.java
BVulaj: Done.

core/src/main/java/com/redhat/lightblue/client/request/BulkLightblueDataRequest.java
BVulaj: I'll work that into this pull request.  I prefer an enum for this case (we have CRUDOperations in core), but if either of you prefer constants instead, just shout.

bserdar: Don't include classes from core.

enum is fine.

BVulaj: Just referencing it as another place we are using enums for this.  Not for actual use here.

I'll commit something shortly.

core/src/main/java/com/redhat/lightblue/client/response/BulkLightblueResponse.java
```diff
@@ -0,0 +1,95 @@
...
+public class BulkLightblueResponse {
+
+    private final List<LightblueResponse> responses;
+    private final List<? super AbstractLightblueRequest> requests;
+
+    private JsonNode json;
+    private String text;
+
+    public BulkLightblueResponse(List<LightblueResponse> resps, AbstractBulkLightblueRequest<? extends AbstractLightblueRequest> reqs) {
+        responses = new ArrayList<LightblueResponse>(resps);
```
bserdar: No need for copy here

core/src/main/java/com/redhat/lightblue/client/response/BulkLightblueResponse.java
```diff
@@ -0,0 +1,95 @@
...
+    public String getText() {
+        if (text == null || text.isEmpty()) {
+            text = getJson().toString();
+        }
+        return text;
+    }
+
+    public JsonNode getJson() {
+        if (json == null || json.toString().isEmpty()) {
+            ObjectNode root = JsonNodeFactory.instance.objectNode();
```
bserdar: json.toString().isEmpty can be very expensive. Just null check is ok

core/src/main/java/com/redhat/lightblue/client/response/BulkLightblueResponse.java
BVulaj: @bserdar This was to support add() since not all List impls support it.  ie - http://docs.oracle.com/javase/6/docs/api/java/util/Arrays.html#asList%28T...%29

I guess we don't really need to support add - I'll refactor slightly.

bserdar: That's the response, why add()?

http/src/main/java/com/redhat/lightblue/client/http/LightblueHttpClient.java
```diff
@@ -272,6 +262,27 @@ public LightblueResponse data(LightblueRequest lightblueRequest) throws Lightblu
     }
 
     @Override
+    public BulkLightblueResponse bulkData(AbstractBulkLightblueRequest<AbstractLightblueDataRequest> request) throws LightblueException {
+        long t1 = new Date().getTime();
```
bserdar: new Date().getTime() -> System.currentTimeMillis()

http/src/main/java/com/redhat/lightblue/client/http/LightblueHttpClient.java
```diff
@@ -272,6 +262,27 @@ public LightblueResponse data(LightblueRequest lightblueRequest) throws Lightblu
...
+        try {
+            String responseBody = httpTransport.executeRequest(request, configuration.getDataServiceURI());
+
+            if (LOGGER.isDebugEnabled()) {
```
bserdar: No need for LOGGER.isDebugEnabled()
For log strings, use "blahblah {}", value, instead of string addition

core/src/main/java/com/redhat/lightblue/client/Operation.java
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * @author bvulaj
+ *
+ */
+public enum Operation {
+    DELETE, FIND, INSERT, SAVE, UPDATE;
+
+    public String toString() {
+        return this.name().toLowerCase();
```
bserdar: No need for toLowerCase(), or toString()

core/src/main/java/com/redhat/lightblue/client/request/AbstractBulkLightblueRequest.java
```diff
@@ -0,0 +1,113 @@
...
+ */
+public abstract class AbstractBulkLightblueRequest<E extends AbstractLightblueRequest> implements LightblueRequest {
+
+    protected List<E> requests;
+
+    public AbstractBulkLightblueRequest() {
+        this(Collections.<E> emptyList());
+    }
+
+    public AbstractBulkLightblueRequest(List<E> requests) {
```
bserdar: No need for this ctor

core/src/main/java/com/redhat/lightblue/client/request/AbstractBulkLightblueRequest.java
```diff
@@ -0,0 +1,113 @@
...
+/**
+ * @author bvulaj
+ *
+ */
+public abstract class AbstractBulkLightblueRequest<E extends AbstractLightblueRequest> implements LightblueRequest {
+
+    protected List<E> requests;
+
+    public AbstractBulkLightblueRequest() {
+        this(Collections.<E> emptyList());
```
bserdar: simply ...=new ArrayList?

core/src/main/java/com/redhat/lightblue/client/request/AbstractBulkLightblueRequest.java
```diff
@@ -0,0 +1,113 @@
...
+
+    /**
+     * Inserts a request after another specified request. This guarantees that the first request parameter will be executed, sequentially, after the second request parameter. It
+     * does not guarantee consecutive execution.
+     * 
+     * @param request
+     * @param after
+     */
+    public void insertAfter(E request, E after) {
+        this.requests.add(this.requests.indexOf(after) + 1, request);
```
bserdar: why this.?

core/src/main/java/com/redhat/lightblue/client/request/data/DataDeleteRequest.java
```diff
@@ -48,4 +48,9 @@ public HttpMethod getHttpMethod() {
     public String getOperationPathParam() {
         return "delete";
     }
+
+    @Override
+    public String getOperation() {
+        return "delete";
```
bserdar: Operation getOperation() ?

BVulaj: Yep.  Just haven't comitted yet.  Incoming.

core/src/main/java/com/redhat/lightblue/client/response/BulkLightblueResponse.java
```diff
@@ -0,0 +1,95 @@
...
+            json = JSON.getDefaultObjectMapper().readTree(responseText);
+            ArrayNode resps = (ArrayNode) json.get("responses");
+            for (Iterator<JsonNode> it = resps.iterator(); it.hasNext();) {
+                JsonNode resp = it.next();
+                LightblueResponse response = new DefaultLightblueResponse(resp.get("response").toString());
+                JsonNode seq = resp.get("seq");
+                if (!seq.isNumber()) {
+                    throw new LightblueException("Invalid sequence.", response);
+                }
+                responses.add(resp.get("seq").intValue(), response);
```
bserdar: This won't work.

Issue #108, opened 2015-05-16
MetadataUpdateSchemaStatusRequest doesn't allow you to set the status or comment


alechenninger:
From spec: `PUT /metadata/{entityName}/{version}/{status}?comment={Change comment}`

Not sure what PUT/metadata/{entityName}/{version}/ will do, but that's the only(ish) request that this class supports currently

PR #159, opened 2015-09-10
Changed MetadataUpdateSchemaStatusRequest to set Comment

snehagunta:
#108  Updated MetadataUpdateSchemaStatusRequest to allow for setting comment and added testcases to get the restURI. Added an appendToURI method to append a query param to restURI.

### DIFF
--- a/core/src/main/java/com/redhat/lightblue/client/request/AbstractLightblueRequest.java
+++ b/core/src/main/java/com/redhat/lightblue/client/request/AbstractLightblueRequest.java
@@ -56,19 +56,18 @@ public abstract class AbstractLightblueRequest implements LightblueRequest {
         restOfURI.append(pathParam);
     }
     
-    protected void appendToURI(StringBuilder restOfURI, String queryParamName, String queryParamvalue) {
+	protected void appendToURI(StringBuilder restOfURI, String queryParamName, String queryParamvalue) {
 		if (!StringUtils.endsWith(restOfURI.toString(), PATH_SEPARATOR)) {
 			if (!StringUtils.contains(restOfURI.toString(), QUERY_PARAM_NAME_VALUE_SEPERATOR)) {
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/redhat/lightblue/client/request/AbstractLightblueRequest.java
```diff
@@ -52,6 +55,24 @@ protected void appendToURI(StringBuilder restOfURI, String pathParam) {
         }
         restOfURI.append(pathParam);
     }
+    
+    protected void appendToURI(StringBuilder restOfURI, String queryParamName, String queryParamvalue) {
+		if (!StringUtils.endsWith(restOfURI.toString(), PATH_SEPARATOR)) {
+			if (!StringUtils.contains(restOfURI.toString(), QUERY_PARAM_NAME_VALUE_SEPERATOR)) {
```
bserdar: Can you simplify this? The if/else blocks share the last three lines, so the if/else can include the first line only.

core/src/main/java/com/redhat/lightblue/client/request/AbstractLightblueRequest.java
```diff
@@ -52,6 +55,23 @@ protected void appendToURI(StringBuilder restOfURI, String pathParam) {
         }
         restOfURI.append(pathParam);
     }
+    
+	protected void appendToURI(StringBuilder restOfURI, String queryParamName, String queryParamvalue) {
+		if (!StringUtils.endsWith(restOfURI.toString(), PATH_SEPARATOR)) {
+			if (!StringUtils.contains(restOfURI.toString(), QUERY_PARAM_NAME_VALUE_SEPERATOR)) {
+				restOfURI.append(QUERY_BEGINNER);
+				restOfURI.append(queryParamName);
```
bserdar: append(queryParamName) can be moved out as well.

Issue #107, opened 2015-05-16
Metadata requests are missing path parameters


alechenninger:
`MetadataCreateSchemaRequest` does not have a way to define schema path param (required)
`MetadataGetEntityNamesRequest` does not have a way to define status (s) path param (optional)

PR #161, opened 2015-09-11
MetadataRequest update

snehagunta:
#107 Updated MetadataCreateSchemaRequest to allow definition of required path param schema.
Updated MetadataGetEntityNamesRequest to allow definition of optional path param status.

### DIFF
--- a/core/src/main/java/com/redhat/lightblue/client/request/metadata/MetadataCreateSchemaRequest.java
+++ b/core/src/main/java/com/redhat/lightblue/client/request/metadata/MetadataCreateSchemaRequest.java
@@ -1,5 +1,7 @@
 package com.redhat.lightblue.client.request.metadata;
 
+import org.apache.commons.lang.StringUtils;
+
 import com.redhat.lightblue.client.http.HttpMethod;
 import com.redhat.lightblue.client.request.AbstractLightblueMetadataRequest;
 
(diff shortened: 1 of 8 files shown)


core/src/main/java/com/redhat/lightblue/client/request/metadata/MetadataCreateSchemaRequest.java
```diff
@@ -5,12 +5,14 @@
 
 public class MetadataCreateSchemaRequest extends AbstractLightblueMetadataRequest {
 
+	private String schema;
```
bserdar: Where is schema added to the submitted request body?

core/src/main/java/com/redhat/lightblue/client/request/metadata/MetadataGetEntityNamesRequest.java
```diff
@@ -1,10 +1,12 @@
 package com.redhat.lightblue.client.request.metadata;
 
+import com.redhat.lightblue.client.enums.MetadataStatus;
 import com.redhat.lightblue.client.http.HttpMethod;
 import com.redhat.lightblue.client.request.AbstractLightblueMetadataRequest;
 
 public class MetadataGetEntityNamesRequest extends AbstractLightblueMetadataRequest {
 
+	private MetadataStatus status;
```
bserdar: How is status added to the URI?

core/src/main/java/com/redhat/lightblue/client/request/metadata/MetadataCreateSchemaRequest.java
```diff
@@ -22,5 +26,23 @@ public String getOperationPathParam() {
...
+
+        requestURI.append(baseServiceURI);
+
+        if (StringUtils.isNotBlank(this.getEntityName())) {
+            appendToURI(requestURI, this.getEntityName());
+        }
+
+    	if (StringUtils.isNotBlank(this.getEntityVersion())) {
+            appendToURI(requestURI, this.schema +"="+ this.getEntityVersion());
+        }
```
bserdar: This won't do. Schema is in request body, not part of uri. Parse string to json, and set the request body.

Also, overload the schema constructor to get a JsonNode schema, because body is json in AbstractListhblueMEtadataRequest

PR #166, opened 2015-09-16
rename DataBulkRequest

dcrissman:


BVulaj:
I'd recommend 'DataBulk' -> 'BulkData' I think.

dcrissman:
> I'd recommend 'DataBulk' -> 'BulkData' I think.

I only did it that way to match existing classes
* DataDeleteRequest
* DataInsertRequest
* etc

This way you type Data and have access to any of the request objects.

PR #171, opened 2015-09-22
Use Literal.values with boxed types instead primitives

ykoer:
Long[] ids = ...
Query.in, Literal.values(ids) 

was causing 
The method values(int...) in the type Literal is not applicable for the arguments (Long[])

### DIFF
--- a/core/src/main/java/com/redhat/lightblue/client/Literal.java
+++ b/core/src/main/java/com/redhat/lightblue/client/Literal.java
@@ -99,14 +99,21 @@ public class Literal extends ExpressionPart implements
         return new Literal(b);
     }
 
-    public static Literal[] values(Integer...v) {
+    public static Literal[] values(int...v) {
         Literal[] ret=new Literal[v.length];
         for(int i=0;i<ret.length;i++)
             ret[i]=Literal.value(v[i]);
         return ret;
     }
 
-    public static Literal[] values(Long...v) {
+    public static Literal[] values(long...v) {
+        Literal[] ret=new Literal[v.length];
+        for(int i=0;i<ret.length;i++)
+            ret[i]=Literal.value(v[i]);
+        return ret;
+    }
+
+    public static Literal[] numbers(Number...v) {
         Literal[] ret=new Literal[v.length];
         for(int i=0;i<ret.length;i++)
             ret[i]=Literal.value(v[i]);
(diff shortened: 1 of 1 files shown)


bserdar:
Does this work with primitive arrays?

ykoer:
right it is not working with primitive arrays.
Looks like varargs and auto-boxing is not working as expected.
See http://stackoverflow.com/questions/2521293/bug-with-varargs-and-overloading
I don't like the conversion path.

bserdar:
How about overloading? Keep the old apis, add the new ones?

ykoer:
I tried that first and it doesn't allow you to use varargs like Literal.values(1,2,3)
Same compile error: The method values(int[]) is ambiguous for the type Literal

bserdar:
Does it work with values(int[]), values(long[]), values(Number[])?

ykoer:
When we just keep
```java
public static Literal[] values(Number...v) {
        Literal[] ret=new Literal[v.length];
        for(int i=0;i<ret.length;i++)
            ret[i]=Literal.value(v[i]);
        return ret;
    }
```
Then we can use arrays but no varargs
```
Literal.value(new long[] {0,1,2});  // OK
Literal.value(new Long[] {0L,1L,2L});  // OK
Literal.value(new int[] {0,1,2});  // OK
Literal.value(new Integer[] {0,1,2});  // OK
Literal.value(new Number[] {0,1,2});  // OK
Literal.value(0,1,3); // The method value(int) in the type Literal is not applicable for the arguments (int, int, int)

alechenninger:
How about Literal.intValues or .ints or .integers or .integerValues where
int uses primitives and integer variants use boxed

alechenninger:
Or you could have values overloads use boxed and have intValues etc for primitives. This is consistent with how Java 8 functional interfaces deal with primitives vs boxed types.

bserdar:
We already have values for primitives. Lets not break things by changing
them.

How about keep existing primitives as they are, and add:

Literal.numbers(Number...n)

ykoer:
we could do that. But do you thing that there are use cases where we wanna use varargs with Literal.values?
Most of the time we have an array or a list?

bserdar:
I can't say which one we have more. We have arrays and lists of both
primitives and boxed values.

PR #184, opened 2015-10-16
Add jmeter test base class

bserdar:


jmeter-tests/pom.xml
```diff
@@ -0,0 +1,62 @@
...
+                     </goals>
+                     <configuration>
+                       <shadedArtifactAttached>true</shadedArtifactAttached>
+                       <shadedClassifierName>alldeps</shadedClassifierName>
+                      <filters>
+                      </filters>
+                    </configuration>
+                  </execution>
+                </executions>
+             </plugin-->
```
dcrissman: If we don't need this xml, can we just remove it?

jmeter-tests/readme.md
```diff
@@ -0,0 +1,25 @@
...
+servo-graphite-0.8.0.jar  (com.netflix.servo:servo-graphite)
+servo-internal-0.8.0.jar   (com.netflix.servo:servo-internal)
+
+Lightblue jars:
+
+jmeter-tests-1.8.0-SNAPSHOT.jar
+lightblue-client-core-1.8.0-SNAPSHOT.jar
+lightblue-client-http-1.8.0-SNAPSHOT.jar
+lightblue-client-hystrix-1.8.0-SNAPSHOT.jar
+lightblue-core-hystrix-1.8.0-SNAPSHOT.jar
```
dcrissman: I am worried that this readme is going to be hard to maintain. I know for sure the lightblue versions are already old. Is the intention that these are the minimum versions required?

jmeter-tests/src/main/java/com/redhat/lightblue/jmeter/AbstractTest.java
```diff
@@ -0,0 +1,83 @@
...
+    public void setupTest(JavaSamplerContext context) {
+        getLogger().debug("Setup test");
+        LightblueHttpClient httpClient;
+        configFile=context.getParameter("ConfigFile",PropertiesLightblueClientConfiguration.DEFAULT_CONFIG_FILE);
+        getLogger().debug("Getting client, config="+configFile);
+        if (configFile != null) {
+            lbClient = new LightblueHttpClient(configFile);
+        } else {
+            lbClient = new LightblueHttpClient();
+        }
```
dcrissman: What about using the LightblueExternalResource junit rule?

Issue #188, opened 2015-11-03
Exception inconsistency


paterczm:
```java
user = lbClient.data(dfr, LBUser.class)
```
will throw LightblueException. However,
```java
LightblueResponse r =  lbClient.data(dfr);
user r.parseProcessed(User.class)
```
will throw both LightblueException and LightblueResponseParseException. I find this behavior inconsistent, because we execute exactly the same steps in both cases (call lightblue, parse json response to create a pojo).

http/src/main/java/com/redhat/lightblue/client/http/LightblueHttpClient.java:
```java
    @Override
    public LightblueResponse data(LightblueRequest lightblueRequest) throws LightblueException {
        LOGGER.debug("Calling data service with lightblueRequest: {}", lightblueRequest.toString());
        return callService(lightblueRequest, configuration.getDataServiceURI());
    }

    @Override
    public <T> T data(AbstractLightblueDataRequest lightblueRequest, Class<T> type) throws LightblueException {
        LightblueResponse response = data(lightblueRequest);
        try {
            return response.parseProcessed(type);
        } catch (RuntimeException | LightblueResponseParseException e) {
            throw new LightblueHttpClientException("Error sending lightblue request: " + lightblueRequest, e);
        }
    }
```
This brings another question: how are LightblueException and LightblueHttpClientException different?

I'm not sure what the best way to clean this up is. Perhaps replace LightblueHttpClientException with LightblueException and make LightblueResponseParseException extend LightblueException?

alechenninger:
IMHO the `data` API should not mangle the `parseProcessed` exceptions, and just remove the try/catch block from `data` entirely. It's weird and surprising that there is this alternate path from `parseProcessed` as you've pointed out. This API should just be sugar, not new behavior. Also, I'd stop throwing `LightblueHttpClientException` except for literally an exceptional situation with the client code itself (not sure where this would actually occur other than maybe an IOException trying to make the request).

Semantically it is important to discern from lightblue had a problem with your request versus the data couldn't be parsed to the model object you passed in, so those two different types (LightblueException and LightblueResponseParseException) of exceptions makes sense at a glance.

paterczm:
[...]

Issue #169, opened 2015-09-17
DefaultLightblueResponse has error API's but a LightblueResponseException is thrown when there is an error


alechenninger:
The API's that return `LightblueResponse` are confusing because of this.

I would either:
1. Remove error API's from LightblueResponse (it becomes a "success lightblue response") and then only return/create it when you have a successful response. The LightblueException contains the error information you need (you will never interact with a response object for successful response if you don't have a successful response).
2. Or do not throw a LightblueException if you are asking for a LightblueResponse. I'm asking the API for an encapsulation of Lightblue's response, which may contain errors, or other information. But you are throwing an exception when that response contains an error, _getting in the way of me just examining Lightblue's response, which is what I asked for, and the API is telling me that it will give me_.

I prefer option 2 but either are better than current state.

The argument to throw an exception when there are errors in the response is to save code and to "error when there is an error." But...
- There is already error information in the response object. There are more ways to show "an error has occurred" than exceptions. Some languages don't even have exceptions at all. So having both is surprising. I would've expected that information to just fall into the response because the API is telling me that's where it will be.
- If you want to save code, you also generally want to parse to a model object, and you will be using an API that does not return a `LightblueResponse`. This scenario is therefore unaffected, and unsurprising, and everyone's happy. Admittedly, this is mainly for find operations, other operations you probably don't care about the response, but still want an exception on error.

alechenninger:
Another option would having two API's, say there's another one like `mandatoryData(....)`, which threw an exception on any error. Or something like this.

You see this in some API's commonly. For instance, Camel has API's for getting message body's which throws an exception if the body is null (`getMandatoryBody()`). But `getBody()` does not.

dcrissman:
Another thing to think about now is BulkLightblueResponse. In it's constructor DefaultLightblueResponse instances are created, so if one of them had an error/data error the entire batch would fail out while parsing the response without finishing other potentially valid responses. We could add a try/catch, but that defeats the purpose of the exception and creates a mess of it's own.

With compounding problem this causes, I am siding with option 2 also.

Couple of thoughts:
* We might then be able to get rid of LightblueResponseException altogether or move it to the Http layer.
* BulkLightblueResponse also throws a LightblueResponseException for an invalid seq id, which we would need to replace. I guess my question there is should a bad seq number, cause the entire batch to fail?

jewzaam:
[...]

PR #189, opened 2015-11-04
Exceptions handling redesigned

paterczm:
This is a breaking change (apis changed). Please don't merge, just say if you're ok merging it.

I think that exception handling in lightblue-client is inconsistent and confusing in general. There are 3 issues pointing at problems with exceptions: #169, #143 and #188. This PR tries to address all of them.

List of exceptions and their usage (in this PR):
* LightblueException, a common parent to all lightblue exceptions
* LightblueHttpClientException, for http exceptions (no response or status != 200)
* LightblueResponseException, for when the response is valid, but contains errors. That means you can get a LightblueResponse object with errors only from this exception (valid response with errors used not cause exception to be thrown)
* LightblueResponseParseException, for json2pojo conversion errors and invalid response structure

In addition to exception related changes, I removed LightblueApacheHttpTransport. It's not used and does not have unit tests.


paterczm:
@ykoer, @bserdar, @alechenninger - I'd like to get your opinions on this.

BVulaj:
If one can only ever get LightblueResponse errors from an Exception, does it still make sense for them to live in the global Response object?  I understand that changing that at all is quite a breaking change though.

core/src/main/java/com/redhat/lightblue/client/response/LightblueResponseException.java
```diff
@@ -0,0 +1,80 @@
...
+ * @author ykoer
+ */
+public class LightblueResponseException extends LightblueException implements LightblueResponseErrorCodes {
+
+    private static final long serialVersionUID = 1L;
+
+    private LightblueResponse lightblueResponse;
+    private Set<String> errorCodes;
+
+    public LightblueResponseException(String message, Throwable cause) {
```
bserdar: Is this ctor still needed? This exception class should have the response.

core/src/main/java/com/redhat/lightblue/client/response/LightblueResponseException.java
```diff
@@ -0,0 +1,80 @@
...
+ * See {@link LightblueResponseErrorCodes} for a list of possible errors.
+ * Example: client can call e.exists(ERR_REST_CRUD_REST_SAVE)
+ *
+ * @author ykoer
+ */
+public class LightblueResponseException extends LightblueException implements LightblueResponseErrorCodes {
+
+    private static final long serialVersionUID = 1L;
+
+    private LightblueResponse lightblueResponse;
```
bserdar: Can this ever be null?

paterczm:
@BVulaj 
> If one can only ever get LightblueResponse errors from an Exception, does it still make sense for them to live in the global Response object? I understand that changing that at all is quite a breaking change though.

The LightblueResponseException contains the LightblueResponse which contains errors. When LightblueResponse is returned normally, the errors are always null - is that what you're referring to? We could have LightblueResponse without error fields and methods and LightblueErrorResponse, extending LightblueResponse, which would have those.

BVulaj:
@paterczm 
> The LightblueResponseException contains the LightblueResponse which contains errors. When LightblueResponse is returned normally, the errors are always null - is that what you're referring to? We could have LightblueResponse without error fields and methods and LightblueErrorResponse, extending LightblueResponse, which would have those.

Yep, that is what I was getting at and suggesting.  If the error field is always going to be null unless wrapped in an exception, I think we should do away with it and extend LightblueResponse

core/src/main/java/com/redhat/lightblue/client/response/LightblueResponseException.java
paterczm: True.

core/src/main/java/com/redhat/lightblue/client/response/LightblueResponseException.java
paterczm: No.

paterczm:
@BVulaj, @bserdar, I incorporated your suggestions (2 commits at the top).

bserdar:
I am fine with this change, but maybe we should release the current, and
inc the version number before merging this?

paterczm:
> I am fine with this change, but maybe we should release the current, and
inc the version number before merging this?

Yes, we could even increase the major version, to indicate it's a breaking change.

jewzaam:
Should `LightblueResponseParseException` extend `LightblueResponseException`?  Or rename to `LightblueParseException`.

+1 to major version change, as this is a breaking change.  Per our standard [1] it must be a major update assuming the client falls into the "application" bucket.

[1] http://docs.lightblue.io/standards/versioning.html

paterczm:
> Should LightblueResponseParseException extend LightblueResponseException? Or rename to LightblueParseException.

LightblueResponseParseException means there are no errors in the response, but parsing failed - either because there is a mismatch between model and metadata or there was a problem parsing the response (client - server mismatch?).

Renamed it to LightblueParseException.

alechenninger:
I argued for @BVulaj 's suggestion already in #169 so +1 (among other options but that was one of them)

ykoer:
There are two kind of problems which we need to address:
1) Business Exceptions, which are expected, but not preventable and therefore reasonable to recover
2) Technical Exceptions, which is everything else. I would throw RuntimeException for those.

The problem here is how do we wanna determine the business errors from so many possible lightblue errors?

paterczm:
LightblueResponseException represents "business" errors, like constraint violation. You need to check error code on that exception for specific error though.

PR #194, opened 2015-11-12
added default value for removeHooks field to allow for default initia…

snehagunta:
…lization of TestController


lightblue-client-integration-test/src/main/java/com/redhat/lightblue/client/integration/test/LightblueExternalResource.java
```diff
@@ -48,7 +48,7 @@ protected AbstractLightblueClientCRUDController getControllerInstance() {
             try {
                 if (removeHooks == Boolean.TRUE)
```
bserdar: removeHooks is not Boolean, it is boolean. So if(removeHooks) { .. } else { ... } is sufficient

PR #206, opened 2015-11-19
remove default constructors from requests as it does not enforce required fields

dcrissman:


BVulaj:
Is there ever a case where we'd want to set something like version after creating the request?

dcrissman:
There shouldn't be. A new request should be created if querying a different entity/version.

PR #208, opened 2015-11-20
fix #200 - add sanity to LightblueErrorResponse

### LINKED ISSUE #200: Review LightblueErrorResponse methods and implementations
This api seems confusing to me. Lightblue errors and data errors mean different things, but I feel the api blurs the distinction.

*hasDataErrors* returns a boolean if data errors exist, however *hasError* returns true if the response has errors or data errors.

The constructor in DefaultLightblueResponse seems confused by this also, as it checks hasError and hasDataErrors, which is a duplication of effort for data errors.

dcrissman:
Having LightblueErrorResponse implies that an error exists on the response, we don't need a method (hasError) to tell us that.

- Remove hasError
- Rename getErrors to getLightblueErrors
- Create getLightblueErrors to return all Error objects


paterczm:
Please don't remove apis from interfaces used in other projects. Deprecate them instead.

dcrissman:
@paterczm methods have been deprecated.

PR #210, opened 2015-12-04
Should always ask for least specific type you need

alechenninger:
Only needs collection, so ask for collection. More or less the same
reason you don't put ArrayList in the parameter. You don't care what the
type is as long as it's a collection.

In consuming code I am trying to pass a map's keySet which is a Set.
Shouldn't have to convert it to a List first.

### DIFF
--- a/core/src/main/java/com/redhat/lightblue/client/request/AbstractDataBulkRequest.java
+++ b/core/src/main/java/com/redhat/lightblue/client/request/AbstractDataBulkRequest.java
@@ -34,12 +34,17 @@ public abstract class AbstractDataBulkRequest<E extends AbstractLightblueRequest
     }
 
     /**
-     * Adds a collection of requests to the end of the current request chain.
+     * Adds a collection of requests to the end of the current request chain in iteration order.
+     *
+     * <p>As the order of responses depends on the order of requests, use a collection with a
+     * reliable iteration order if you need to get a specific response for a specific request by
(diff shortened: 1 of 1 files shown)


bserdar:
Bulk requests are ordered. Collection is not.

alechenninger:
Does that really matter in this case though? You're still adding it to a List, and whether or not the caller cares about the response List being in the same order as the Collection they added is up to them. If they care, they can still use a List. In the response object you can get an individual response from a specific request instance which figures out the index of that response, so when using that API I don't have to know what the order of the requests was. Am I missing something?

Thanks for taking a look!

bserdar:
I suppose that's ok. You could add a comment to that api saying if the
input collection is ordered, the output will match that, but if not, there
is no guarantee on the order of the output.

alechenninger:
I think build failure is was intermittent problem with integration tests, not related to my change. Builds fine locally (and the openjdk travis build went fine too). Could you try kicking off another travis build to confirm?

dcrissman:
@alechenninger I have restarted the job several times now and keep getting red lights. Locally it passes for me with both java 7 and 8. I rebuilt the master branch, and that green lights. I am in support of this change, but want to understand the error before merging.

dcrissman:
Other recent PRs have built just fine.

dcrissman:
@alechenninger merge in changes from master, it might fix the issue.

alechenninger:
@dcrissman Fixed

PR #212, opened 2015-12-08
jmeter base class

bserdar:


jmeter-tests/src/main/java/com/redhat/lightblue/jmeter/AbstractTest.java
```diff
@@ -0,0 +1,83 @@
...
+        LightblueHttpClient httpClient;
+        configFile=context.getParameter("ConfigFile",PropertiesLightblueClientConfiguration.DEFAULT_CONFIG_FILE);
+        getLogger().debug("Getting client, config="+configFile);
+        if (configFile != null) {
+            lbClient = new LightblueHttpClient(configFile);
+        } else {
+            lbClient = new LightblueHttpClient();
+        }
+        getLogger().debug("lbClient="+lbClient);
+        name = context.getParameter(TestElement.NAME);
```
dcrissman: What about using the LightblueExternalResource junit rule to stand a lightblue instance?

bserdar: The point of this is to call a remote lightblue cluster.

dcrissman:
@bserdar looks like travis is failing because you are relying on an older version of the lightblue client. Would you please bump that to the most recent?

bserdar:
Done

jmeter-tests/src/main/java/com/redhat/lightblue/jmeter/AbstractTest.java
dcrissman: Ok, so the intention is that implementing test classes would provide their own lightblue cluster.

bserdar: Yes, you have to have a running lightblue somewhere to use these jmeter
tests.

PR #220, opened 2015-12-15
Updated core version

jewzaam:
Many projects depending on this client do not pull in specific versions of lightblue-core.  Version 1.9.0 has some bugs, this PR is to make sure when next client version is cut we get an updated dependency on core.


jewzaam:
Or should we make these kinds of dependencies provided, forcing the inclusion of specific versions of artifacts like core if needed by the client?

dcrissman:
I will merge this PR. I think ideally the client should include the lightblue libraries that is required by the client application to talk to lightblue.

PR #221, opened 2015-12-16
Use any available port in wiremock tests

alechenninger:
Curious if this fixes travis issues


alechenninger:
@dcrissman This seemed to do the trick!

PR #227, opened 2016-01-05
For bulk data response, there is no need to parse, toString, and then parse the json again

dcrissman:


core/src/main/java/com/redhat/lightblue/client/response/AbstractLightblueResponse.java
```diff
@@ -30,6 +30,16 @@ public AbstractLightblueResponse(String responseText, ObjectMapper mapper) throw
         }
     }
 
+    public AbstractLightblueResponse(JsonNode responseNode) {
+        this(responseNode, JSON.getDefaultObjectMapper());
+    }
+
+    public AbstractLightblueResponse(JsonNode responseNode, ObjectMapper mapper) {
```
bserdar: Is there any point passing this mapper around?

dcrissman: I would love to say no, but DefaultLightblueDataResponse#parseProcessed uses it, so it needs to be there. The intention is that the default will always be used, unless a custom one is provided, in which case the custom one will always be used.

bserdar:
It says conflicts with base, needs rebase

dcrissman:
@bserdar base merged in

Issue #202, opened 2015-11-19
Bulk Data Response with errored Data Responses does not process remaining responses


dcrissman:
DefaultLightblueBulkDataResponse iterates over the raw json, but passes the individual response to DefaultLightblueResponse to parse. If a DefaultLightblueResponse throws an exception while parsing, then the entire DefaultLightblueBulkDataResponse stops processing additional DefaultLightblueResponses. It is possible that there were successful DefaultLightblueResponse inter-mixed with the failed ones.

I feel that the BulkLightblueResponse should be constructed and returned with all responses on it, both failed and successful, regardless of exceptions. Let the client application figure out what to do with the failures.

alechenninger:
What's the status on this? This is important to the event handler work we're doing since we're using bulk requests heavily.

dcrissman:
This is not actively being worked on at this time as there are other higher priorities for us.

alechenninger:
Alright, if I understand correctly, the parsing errors that would happen here would only be if the client had an incorrect understanding of the response coming from lightblue, which would be a lightblue or client bug and not a bug of a consumer of the client library?

dcrissman:
Not sure what you mean exactly, but take the following bulk response for example:
response a. good
response b. error
response c. good
response d. error

In the above example, all you would get is a LightblueErrorResponse for b, with no mention of a, c, or d.

https://github.com/lightblue-platform/lightblue-client/blob/master/core/src/main/java/com/redhat/lightblue/client/response/DefaultLightblueBulkDataResponse.java#L40

alechenninger:
Oh I see. `DefaultLightblueDataResponse` throws an exception if the response contains any kind of error. That's pretty bad for us. Can you prioritize fixing this?

dcrissman:
@alechenninger I haven't formally created a PR yet because I want to write some unit tests first. That said, I would appreciate any feedback you may have. See my earlier commit.

dcrissman:
@bserdar @jewzaam @BVulaj @paterczm @snehagunta - I would appreciate any of your input also.

Essentially what I am purposing is a breaking change to DefaultLightblueBulkDataResponse, that will allow all of the individual DefaultLightblueDataResponse to be parsed even if one/multiple have an 'error' or 'dataerror' on it, instead of failing the entire response and the client getting nothing.

This is consistent with the current DefaultLightblueDataResponse implementation of throwing exceptions when a 'error' or 'dataerror' is returned from lightblue.

[...]

PR #230, opened 2016-01-08
Issue #202 - Bulk Data Response now returns all responses reguardless of any lb errors

dcrissman:
See #202 for conversation about this fix.

This is a breaking change as DefaultLightblueBulkDataResponse now throws LightblueBulkResponseException instead of LightblueResponseException.

Also, seq exception now throws a LightblueParseException.

### DIFF
--- a/core/src/main/java/com/redhat/lightblue/client/response/LightblueBulkResponseException.java
+++ b/core/src/main/java/com/redhat/lightblue/client/response/LightblueBulkResponseException.java
@@ -1,50 +1,29 @@
 package com.redhat.lightblue.client.response;
 
-import java.util.ArrayList;
 import java.util.Collections;
-import java.util.HashSet;
-import java.util.List;
 import java.util.Map;
-import java.util.Set;
-import java.util.SortedMap;
-import java.util.TreeMap;
 
 public class LightblueBulkResponseException extends LightblueException {
 
     private static final long serialVersionUID = -1204842853642316889L;
 
-    private final LightblueBulkDataResponse bulkResponse;
-    private final Map<Integer, LightblueResponseException> erroredResponses;
+    private final LightblueBulkDataErrorResponse bulkResponse;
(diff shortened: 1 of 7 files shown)


dcrissman:
@alechenninger @BVulaj - Here is what I came up with. Let me know what you think.

core/src/main/java/com/redhat/lightblue/client/response/LightblueBulkResponseException.java
```diff
@@ -0,0 +1,29 @@
...
+    public LightblueBulkResponseException(String message,
+            LightblueBulkDataResponse bulkResponse,
+            Map<Integer, LightblueResponseException> erroredResponses) {
+        super(message);
+        this.bulkResponse = bulkResponse;
+        this.erroredResponses = erroredResponses;
+    }
+
+    public LightblueBulkDataResponse getBulkResponse() {
+        return bulkResponse;
```
BVulaj: Do you think there is value in doing this rather than returning a collection of the successful responses?

BVulaj:
Looks good to me. I just had the one comment that I don't feel too strongly about either way.

core/src/main/java/com/redhat/lightblue/client/response/LightblueBulkResponseException.java
dcrissman: I did this to be consistent with how LightblueResponseException works.

That said, I think a unit test showing how to get just the successful responses might be useful. Then, maybe it could be made easier.

BVulaj: I'll vote for that.  If you think that'll be straight forward let's get it into this PR and then I'll merge.

dcrissman:
Changes made.

As this is a breaking change, the release version should also be bumped. I will submit a separate PR after this is merged.

Issue #128, opened 2015-06-21
It is not clear that file:// prefix is required for system paths to certificate files


paterczm:
If you want to load certificates from a system path, you need to use file:// prefix in the lightblue client configuration file. Omitting the prefix will result in ```java.security.cert.CertificateException: Missing input stream``` exceptions, which will make you check the paths, filesystem permissions, etc. and generally waste your time. The exception message needs to be updated with a short explanation of this behavior.

Related to https://github.com/lightblue-platform/lightblue-client/commit/4a6932b307fc1f061e88477a01dbf3d14e3414a0.

PR #233, opened 2016-01-11
Add javadocs for client cert locations + more descriptive errors

cainj13:
The method names for the private/CA cert paths don't really denote that they're supposed to be on the classpath, thought it might be helpful to make that a little more clear (tripped me up for a moment anyway).

Use the Java 7 Objects class to check for null value - let me know if LB has a standard way of doing this or isn't on 7.  Didn't see a source/target java compile version in the parent pom, would be happy to add one though.

### DIFF
--- a/core/src/main/java/com/redhat/lightblue/client/LightblueClientConfiguration.java
+++ b/core/src/main/java/com/redhat/lightblue/client/LightblueClientConfiguration.java
@@ -72,30 +72,34 @@ public class LightblueClientConfiguration {
     }
 
     /**
-     * @return classpath reference for the CA file used to authenticate the validity of
-     *         the LB server certificate
+     * @return Reference for the CA file used to authenticate the validity of
+     *         the LB server certificate.  Should either be on the classpath, or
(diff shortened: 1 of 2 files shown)


dcrissman:
Thanks for this PR and sorry you got hung up on this issue.

This actually looks very similar to #128. Appending file:// to the paths should fix your issue. Let me know if that addresses your problem or not. If it does, then I can look into writing a better exception message.

> Didn't see a source/target java compile version in the parent pom

Yes, the client is using 1.7. The source/target are defined in the parent pom "lightblue-client-pom".

cainj13:
Gotcha, I see where you're doing that now in SslSocketFactories.  I can adjust the error message/javadocs appropriately, looks like I'm not the only one who went on a wild goose chase because of this.  Would that be more useful?

> Yes, the client is using 1.7. The source/target are defined in the parent pom "lightblue-client-pom".

doh!  must've missed them before.

dcrissman:
Go ahead and make the changes. Thanks!

PR #234, opened 2016-01-11
Fixes #224 - Locking api should not swallow exceptions

### LINKED ISSUE #224: Client swallows lightblue errors when trying to acquire lock
If lightblue responds with an error, acquire simply returns false. It should throw an exception.

dcrissman:
See #224


dcrissman:
@alechenninger - I assume this would resolve your concern.

http/src/main/java/com/redhat/lightblue/client/http/LightblueHttpClient.java
```diff
@@ -206,7 +205,7 @@ public Locking getLocking(String domain) {
      * @see com.redhat.lightblue.client.LightblueClient#metadata(com.redhat.lightblue .client.request.LightblueRequest)
      */
     @Override
-    public DefaultLightblueMetadataResponse metadata(LightblueRequest lightblueRequest) throws LightblueException {
+    public DefaultLightblueMetadataResponse metadata(LightblueRequest lightblueRequest) throws LightblueParseException, LightblueResponseException, LightblueHttpClientException {
```
paterczm: Why not use LightblueException? In most cases you don't care why it failed.

core/src/main/java/com/redhat/lightblue/client/response/DefaultLightblueSingleValueResponse.java
```diff
@@ -0,0 +1,133 @@
...
+import com.redhat.lightblue.client.util.JSON;
+
+/**
+ * <p>A special response type for responses from lightblue where only a single primitive value is expected.</p>
+ * <p><b>WARNING:</b> There should be only a few special cases where this response type is needed. It is not
+ * recommended for general usage.</p>
+ *
+ * @author dcrissman
+ */
+public class DefaultLightblueSingleValueResponse extends AbstractLightblueResponse implements LightblueResponse, LightblueErrorResponse {
```
paterczm: This class contains redundant code. Why not extend DefaultLightblueDataResponse?

core/src/main/java/com/redhat/lightblue/client/response/DefaultLightblueDataResponse.java
```diff
@@ -51,30 +51,6 @@ public boolean hasDataErrors() {
         return err != null && !(err instanceof NullNode) && err.size() > 0;
     }
 
-    @Deprecated
-    @Override
-    public boolean hasError() {
```
paterczm: This change is being made long after version bump, a little concerned it will break the build.

http/src/main/java/com/redhat/lightblue/client/http/LightblueHttpClient.java
dcrissman: The client side try/catch block can still catch the generic LightblueException. Why not provide the ability to handle individually if desired?

core/src/main/java/com/redhat/lightblue/client/response/DefaultLightblueSingleValueResponse.java
dcrissman: Yes, this is duplicate logic. I did this because DefaultLightblueSingleValueResponse is not a LightblueDataResponse. But you are right for the hand slap, let me see if there is a way I can abstract the logic so it is not duplicated.

core/src/main/java/com/redhat/lightblue/client/response/DefaultLightblueDataResponse.java
dcrissman: We had to break the build already for #230. The major version was bumped in #232.

core/src/main/java/com/redhat/lightblue/client/response/DefaultLightblueSingleValueResponse.java
paterczm: Maybe just add a new method to get single value from response instead of creating a whole new class?

dcrissman: The appropriate exception is LightblueResponseException, and that requires an instance of LightblueErrorResponse.

dcrissman: Logic extracted to DefaultLightblueErrorResponse

lightblue-client-integration-test/src/test/java/com/redhat/lightblue/client/integration/test/LockingTest.java
```diff
@@ -32,13 +36,43 @@ public void testAcquireAndRelease() throws Exception {
...
+
+        //assert the lock was successfully released.
+        try {
+            lbLocker.getLockCount("test-lock");
+            fail();
+        } catch (LightblueResponseException e) {
+            Error error = e.getLightblueResponse().getLightblueErrors()[0];
+            assertNotNull(error);
+            assertTrue(error.getMsg().contains("InvalidLockException: test-lock"));
+        }
```
dcrissman: I mostly just want to highlight this consequence, but here is a case where not swallowing the exception makes for uglier client side code.

dcrissman: @bserdar - Should a get lock count return an error from lightblue if the resourceId is unknown? Or could it just return 0?

bserdar: Hm, it is a bit misleading I think, but reasonable. However, the convention
for all the locking APIs is that if the lock is invalid, and a lock becomes
invalid once it is unlocked, they throw exceptions. If it is changed here,
it should be changed in others as well.

One way to deal with ugly code is to create separate lock exceptions for
client, and to make them runtime exceptions.

bserdar: Lock exceptions are not exactly response exceptions. Maybe you can consider
using a different base exception, something like LockException?

dcrissman: I think the argument could be made that a lock exception is a response exception. It is just a error on a response from lightblue that happens to be for a lock.

bserdar: I don't see it that way, but that's just semantics. For sake of clarity, I
think lock exceptions can be thought completely separate from response
errors. That makes client code much cleaner. I don't see any benefit in
extending a common exception for this case. Lock errors don't contain any
diagnostic information.

dcrissman: Made the changes as discussed, this call is now simplified

dcrissman:
All requested changes should be made at this point. Let me know if you have any feedback.

dcrissman:
I created #247 to sort out on what level data errors should be handled as bulk, metadata, and lock do not have them.

core/src/main/java/com/redhat/lightblue/client/response/DefaultLightblueDataResponse.java
```diff
@@ -157,4 +77,12 @@ public JsonNode getProcessed() {
         }
     }
 
+    protected int parseInt(String fieldName) {
+        JsonNode field = getJson().findValue(fieldName);
+        if (field == null || field.isNull()) {
```
bserdar: findValue(x) searches x recursively in the node, or its children. Is this really what you want here? I think you should use get(x).

dcrissman: Good catch. fixed.

dcrissman:
Due to the breaking changes introduced, I created https://github.com/lightblue-platform/lightblue-client/pull/253 to bump the major version.

PR #241, opened 2016-01-12
fixes #229 - Support try-with-resources syntax for locking

### LINKED ISSUE #229: Support try-with-resources syntax for locking
`com.redhat.lightblue.client.Locking` or similar could produce `java.io.Closeable` instances.

Then locks could be used in try-with-resources statements, which would be convenient.

dcrissman:
#229


core/src/main/java/com/redhat/lightblue/client/Locking.java
```diff
@@ -49,4 +52,40 @@ public int getLockCount(String resourceId) throws LightblueException {
...
+        public int getLockCount() throws LightblueException {
+            return Locking.this.getLockCount(resourceId);
+        }
+
+        @Override
+        public void close() throws IOException {
+            try {
+                release(resourceId);
+            } catch (LightblueException e) {
+                throw new IOException("Unable to release lock: " + resourceId, e);
```
dcrissman: I am still debating about this re-throw as an IOException. Everywhere else the LB exception bubble up, so this is a notable deviation from that. Additionally, I don't think of the LB exception as a IOException. The alternative is to use a RuntimeException, but that doesn't solve the first issue and is now unchecked.

core/src/main/java/com/redhat/lightblue/client/Locking.java
```diff
@@ -49,4 +52,40 @@ public int getLockCount(String resourceId) throws LightblueException {
     public boolean ping(String resourceId) throws LightblueException {
         return ping(callerId,resourceId);
     }
+
+    public Lock lock(String resourceId) throws LightblueException {
+        return lock(resourceId, null);
+    }
+
+    public Lock lock(String resourceId, Long ttl) throws LightblueException {
+        acquire(resourceId, ttl);
```
bserdar: This isn't right. It shouldn't return a lock if acquire returns false.

core/src/main/java/com/redhat/lightblue/client/Locking.java
bserdar: IOException comes from API signature, so there's nothing we can do there. You can add a release() with the LightblueException.

dcrissman:
@bserdar changes made.

bserdar:
Please rebase from master

dcrissman:
@bserdar merged

PR #248, opened 2016-01-13
remove bad classes and deprecate one of them

dcrissman:
If the goal is to push the use of our apis to build requests, then these classes stand in direct opposition to that goal. I am removing them. DataInsertRequestStub is being kept for now as it is still being used internally, with the intention of removing it also.


dcrissman:
Created https://github.com/lightblue-platform/lightblue-client/issues/249 to remove the remaining class.

PR #268, opened 2016-03-04
Add support for basic http auth

alechenninger:
I'm not crazy about how cert auth and basic auth info can live simultaneously as peers, it's a little weird, but it works.

I cleaned up few things nearby, boy scout style.


http/src/main/java/com/redhat/lightblue/client/http/transport/JavaNetHttpTransport.java
```diff
@@ -126,8 +149,6 @@ public String executeRequest(LightblueRequest request, String baseUri) throws Li
             }
 
             return response(connection);
-        } catch (ProtocolException e) {
-            throw new LightblueHttpClientException(e);
```
alechenninger: Since ProtocolException is an IOException, and they both do the same thing, I thought it was cleaner / DRY to remove this block.

http/src/test/java/com/redhat/lightblue/client/http/transport/JavaNetHttpTransportTest.java
```diff
@@ -287,17 +302,4 @@ public T answer(InvocationOnMock invocation) throws Throwable {
             return returnValue;
         }
     }
-
-    static class FakeConnectionFactory implements JavaNetHttpTransport.ConnectionFactory {
```
alechenninger: This was not used anywhere

dcrissman:
It might be useful to make it clear that you should not be using both cert auth and basic auth. Maybe throw an exception if both exist?

alechenninger:
I think it might have more annoyance than value to fail at runtime with that scenario. If you provide both, it doesn't hurt. I could see a refactoring that enforced this at compile time if you were building your config using Java (where config accepted a single AuthMethod type of some sort which was _either_ basic or cert-based). But, that is a larger refactoring that necessary I think for now IMHO.

dcrissman:
I have no objections to this change. It might be worth documenting which method wins if both are provided.

Needs a rebase.

dcrissman:
This is needed in order to complete https://github.com/lightblue-platform/lightblue-rest/issues/208.

alechenninger:
@dcrissman Merged master.

Issue #267, opened 2016-03-02
Remove deps on lightblue core and rest!!!!!


jewzaam:
This is leading to dependency hell for releases.  Client shouldn't require either of these!  We're looking at releasing only 4.0.0 now and it is dependent on changes in rest at least...

paterczm:
lightblue-rest is a dependency because it's used in integration testing (lightblue-client-integration-test). lightblue-client-integration-test used to be standalone, but at some point it was added to the client.

paterczm:
And lightblue-core-hystrix dependency is used because client uses hystrix. Since we're considering to drop hystrix altogether, we can start by removing it from the client.

paterczm:
@dcrissman, fyi.

dcrissman:
com.redhat.lightblue.rest:lightblue-rest-integration-test makes LightblueClientTestHarness possible

lightblue-client-integration-test was added to the lightblue-client because it makes sense that the client would include tools for testing using the client.

As for hystrix, I have no objections to it being removed if we are removing hystrix anyway.

jewzaam:
Re lightblue-client-integration-test, I'd recommend breaking into a
separate repo so we don't tie client to rest.  We can incorporate it in the
lightblue repo as a submodule and have it built every time that is built
(nightly internally and when touched via travis).

Hystrix we had talked about breaking out.  Thought we were considering
keeping it for client only.  If so, hystrix could move to client?

dcrissman:
> Re lightblue-client-integration-test, I'd recommend breaking into a separate repo so we don't tie client to rest. We can incorporate it in the lightblue repo as a submodule and have it built every time that is built (nightly internally and when touched via travis).

This change would also force us to move integration testing, or testing in the client that is based on the LightblueClientTestHarness, to this external project also.

What dependency hell were you running into? I assume you wanted to release an application using the new version of the client without releasing the rest layer, but I think that is somewhat problematic as well as it assumes there were no breaking changes in the rest layer.

jewzaam:
Wanted to release the client without releasing core and rest.  There isn't
a good reason in my mind to have these dependencies with a JSON based REST
interface.  There's no code to share.  Client doesn't need to even do
validation.  If there's a change in rest then yes client needs rest to be
released, but that's not a build-time dependency.

dcrissman:
[...]

PR #277, opened 2016-03-17
fixes #273 - remove hard dependency on lightblue-core-hystrix

### LINKED ISSUE #273: Remove lightblue-core-hystrix as a dependency for the client
This ticket is split off from #267 from the hystrix elements.

The LightblueHystrixClient has a dependency on lightblue-core in order to use the ServoGraphiteSetup utiliy class. To resolve this issue, we either need to split the class in core into it's own library (that supports jdk7) or copy it into lightblue-client.

dcrissman:


dcrissman:
This still fails, but actually builds further than it was before. So the fix work, but #267 needs to be resolved before the build will be totally stable again. Merge if you want or wait until everything is fixed.

jewzaam:
jdk7 build fails.  Can a profile be setup for jdk7 that doesn't include the integration tests?

bserdar:
Can we split this into two projects, one for 1.7 and one for 1.8, similar
to the structure we have for lightblue-rest?

jewzaam:
How about a profile for jdk7 vs jdk8 where the difference is the modules built?  Is there a reason to bother building integration tests with jdk7?
 
http://maven.apache.org/guides/introduction/introduction-to-profiles.html#Profiles_in_POMs

dcrissman:
Perhaps not, but again this will lock a jdk7 application at a certain
version. Probably ok, but doesn't resolve bug fixes or other updates.

jewzaam:
It only means integration tests would run with jdk8, not jdk7.  If you want
to run them, even if your client needs jdk7, you build with jdk8.  It isn't
something used at runtime, so wouldn't that be reasonable?

jewzaam:
And the client jar's published would be jdk7

dcrissman:
I don't believe this conversation is actually being recorded in github. We
should move to either mojo or github.

dcrissman:
Discussion did get logged to github, but to wrong issue. Above conversation should be on https://github.com/lightblue-platform/lightblue-client/issues/267

PR #282, opened 2016-04-07
Revert and fix "Removing hystrix b/c it breaks java7 build"

dcrissman:
Working PR to see why hystrix is failing.


dcrissman:
Ready for review.

PR #283, opened 2016-04-07
integration tests should still compile with java7, just only run tests with java8

dcrissman:


dcrissman:
No, I take that back. Closing.

PR #292, opened 2016-05-04
Add support for generator api

bserdar:


core/src/main/java/com/redhat/lightblue/client/request/data/GenerateRequest.java
```diff
@@ -0,0 +1,98 @@
...
+ * Parse the results as:
+ * <pre>
+ *	String[] results = client.data(request, String[].class);
+ * </pre>
+ * 
+ */
+public class GenerateRequest extends AbstractLightblueDataRequest {
+
+    private Integer n;
+    private String path;
```
jewzaam: Is there a way to make path required?  Clients could create this using constructor without path then try to make the request without calling path(...)

bserdar: Only if I remove the path() function, and use this as an immutable class
with only ctors. But then i'll lose the ability to have ctors for default
versions, because that'll be ambiguous. Still, can be done. Should I?

That still doesn't prevent passing in null for path to ctor though.,

bserdar: What I said was wrong. We can still have default versions with immutable
class. Should I make that?

jewzaam: Given null could be passed to constructor it's not going to fix anything.  If anything have getRestURI fail if there's no path?  Would save a request to lightblue if it was malformed.

bserdar: Added null check.

PR #293, opened 2016-05-05
Prevent generator api from being used in bulk request

bserdar:


core/src/main/java/com/redhat/lightblue/client/LightblueClient.java
```diff
@@ -17,6 +17,6 @@
 
     LightblueBulkDataResponse bulkData(AbstractDataBulkRequest<AbstractLightblueDataRequest> requests) throws LightblueException;
 
-	<T> T data(AbstractLightblueDataRequest lightblueRequest, Class<T> type) throws LightblueException;
+	<T> T data(LightblueRequest lightblueRequest, Class<T> type) throws LightblueException;
```
dcrissman: Why?

dcrissman: I feel strongly that the data api should be locked down to AbstractLightblueDataRequest implementations. The method is not designed to work with other types of implementations.

dcrissman:
What if we created a marker interface that extended LightblueRequest for requests that can be bulk packaged? Say WithBulkSupport or something? Then the DataBulkRequest could only accept only requests with that interface?

core/src/main/java/com/redhat/lightblue/client/LightblueClient.java
bserdar: I don't agree. Then there should be two interfaces, and two methods. Having
an interface, and then requiring the abstract implementation of it is ugly.

bserdar:
+1

core/src/main/java/com/redhat/lightblue/client/LightblueClient.java
dcrissman: AbstractLightblueDataRequest is the parent of all the data request objects implemented in the lightblue client. Specifically it knows how to build the rest uri.

core/src/main/java/com/redhat/lightblue/client/request/AbstractLightblueMetadataRequest.java
```diff
@@ -36,14 +54,7 @@ public String getRestURI(String baseServiceURI) {
     }
 
     @Override
-    public String getBody() {
-        JsonNode body = getBodyJson();
-        if (body == null) {
-            return null;
-        }
-        return body.toString();
+    public JsonNode getBodyJson() {
```
dcrissman: abstract?

core/src/main/java/com/redhat/lightblue/client/request/AbstractLightblueMetadataRequest.java
```diff
@@ -4,30 +4,48 @@
 
 import com.fasterxml.jackson.databind.JsonNode;
 
-public abstract class AbstractLightblueMetadataRequest extends AbstractLightblueRequest {
+import com.redhat.lightblue.client.http.HttpMethod;
 
-    public AbstractLightblueMetadataRequest(String entityName, String entityVersion) {
-        super(entityName, entityVersion);
+public abstract class AbstractLightblueMetadataRequest extends LightblueRequest {
```
dcrissman: Why is this class's name not changed?

bserdar: Changed now,

core/src/main/java/com/redhat/lightblue/client/request/AbstractLightblueMetadataRequest.java
bserdar: abstract, instead of return null? could be, but then I have to implement it
in all the concrete classes.

dcrissman:
Not merging per @bserdar's request, but +1 to this change.

bserdar:
@BVulaj , please take a look at the changes to bulk requests

paterczm:
I like the changes. Since there are compatibility concerns, please remember to bump the version, even if it's SNAPSHOT (not seeing version bump in the diff).

bserdar:
Version bumped.

PR #297, opened 2016-05-13
Adding execution to request with some utilities to create MongoContro…

paterczm:
…ller options

### DIFF
--- a/core/src/main/java/com/redhat/lightblue/client/Execution.java
+++ b/core/src/main/java/com/redhat/lightblue/client/Execution.java
@@ -1,0 +1,46 @@
+package com.redhat.lightblue.client;
+
+import com.fasterxml.jackson.databind.node.JsonNodeFactory;
+
+/**
+ * https://jewzaam.gitbooks.io/lightblue-specifications/content/language_specification/execution.html
+ *
+ * @author mpatercz
+ *
+ */
+public class Execution extends Expression {
+
+    public static class MongoController {
+
+        public static final String READ_PREFERENCE = "readPreference",
+                WRITE_CONCERN = "writeConcern",
+                MAX_QUERY_TIME_MS = "maxQueryTimeMS";
+
+        public enum ReadPreference {
+            nearest, primaryPreferred, secondary, secondaryPreferred;
+        }
+
+        public static Execution withReadPreference(ReadPreference readPreference) {
+            Execution e = new Execution();
+            e.add(READ_PREFERENCE, readPreference.toString());
+            return e;
+        }
+
+        public static Execution withWriteConcern(String writeConcern) {
+            Execution e = new Execution();
+            e.add(WRITE_CONCERN, writeConcern);
+            return e;
+        }
+
+        public static Execution withMaxQueryTime(int maxQueryTimeMS) {
+            Execution e = new Execution();
+            e.add(MAX_QUERY_TIME_MS, JsonNodeFactory.instance.numberNode(maxQueryTimeMS));
+            return e;
+        }
+    }
+
+    public Execution() {
+        super(false);
+    }
+
+}
--- a/core/src/main/java/com/redhat/lightblue/client/Execution.java
+++ b/core/src/main/java/com/redhat/lightblue/client/Execution.java
@@ -1,6 +1,7 @@
 package com.redhat.lightblue.client;
 
 import com.fasterxml.jackson.databind.node.JsonNodeFactory;
+import com.redhat.lightblue.client.Execution.MongoController.ReadPreference;
 
 /**
  * https://jewzaam.gitbooks.io/lightblue-specifications/content/language_specification/execution.html
@@ -17,29 +18,38 @@ public class Execution extends Expression {
                 MAX_QUERY_TIME_MS = "maxQueryTimeMS";
 
         public enum ReadPreference {
-            nearest, primaryPreferred, secondary, secondaryPreferred;
+            nearest, primaryPreferred, primary, secondary, secondaryPreferred;
         }
 
         public static Execution withReadPreference(ReadPreference readPreference) {
-            Execution e = new Execution();
-            e.add(READ_PREFERENCE, readPreference.toString());
-            return e;
+            return new Execution().addReadPreference(readPreference);
         }
 
         public static Execution withWriteConcern(String writeConcern) {
-            Execution e = new Execution();
-            e.add(WRITE_CONCERN, writeConcern);
-            return e;
+            return new Execution().addWriteConcern(writeConcern);
         }
 
-        public static Execution withMaxQueryTime(int maxQueryTimeMS) {
-            Execution e = new Execution();
-            e.add(MAX_QUERY_TIME_MS, JsonNodeFactory.instance.numberNode(maxQueryTimeMS));
-            return e;
+        public static Execution withMaxQueryTimeMS(int maxQueryTimeMS) {
+            return new Execution().addMaxQueryTimeMS(maxQueryTimeMS);
         }
     }
 
-    public Execution() {
+    public Execution addReadPreference(ReadPreference readPreference) {
+        this.add(MongoController.READ_PREFERENCE, readPreference.toString());
+        return this;
+    }
+
+    public Execution addWriteConcern(String writeConcern) {
+        this.add(MongoController.WRITE_CONCERN, writeConcern);
+        return this;
+    }
+
+    public Execution addMaxQueryTimeMS(int maxQueryTimeMS) {
+        this.add(MongoController.MAX_QUERY_TIME_MS, JsonNodeFactory.instance.numberNode(maxQueryTimeMS));
+        return this;
+    }
+
+    private Execution() {
         super(false);
     }
 
--- a/core/src/main/java/com/redhat/lightblue/client/request/AbstractLightblueDataExecutionRequest.java
+++ b/core/src/main/java/com/redhat/lightblue/client/request/AbstractLightblueDataExecutionRequest.java
@@ -1,0 +1,28 @@
+package com.redhat.lightblue.client.request;
+
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.redhat.lightblue.client.Execution;
+
+public abstract class AbstractLightblueDataExecutionRequest extends AbstractLightblueDataRequest {
+
+    public AbstractLightblueDataExecutionRequest(String entityName, String entityVersion) {
+        super(entityName, entityVersion);
+    }
+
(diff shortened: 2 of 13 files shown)


core/src/main/java/com/redhat/lightblue/client/request/data/DataDeleteRequest.java
```diff
@@ -31,6 +31,7 @@ public JsonNode getBodyJson() {
         if (queryExpression != null) {
             node.set("query", queryExpression.toJson());
         }
+        appendExecutionToJson(node);
```
bserdar: It might be better if getBodyJson is implemented in abstractLightblueDataRequest as a method that creates a json node, and adds execution to it, and in all the subclasses, the overriding getBodyJson would first call super.getBodyJson(), and then add its bits to it.

core/src/main/java/com/redhat/lightblue/client/Execution.java
```diff
@@ -0,0 +1,46 @@
...
+            return e;
+        }
+
+        public static Execution withWriteConcern(String writeConcern) {
+            Execution e = new Execution();
+            e.add(WRITE_CONCERN, writeConcern);
+            return e;
+        }
+
+        public static Execution withMaxQueryTime(int maxQueryTimeMS) {
```
jewzaam: Would rename to withMaxQueryTimeMS so it maps to documented option and unit is clearer.

core/src/main/java/com/redhat/lightblue/client/request/data/DataDeleteRequest.java
jewzaam: It's done similar to how range was implemented.  Should clean up both if this change is implemented.

core/src/test/java/com/redhat/lightblue/client/request/TestDataInsertRequest.java
```diff
@@ -31,4 +32,14 @@ public void testProjectionsAsArray() {
                 "\"projection\":{\"field\":\"*\",\"include\":true,\"recursive\":false}"));
     }
 
+    @Test
+    public void testExecutionReadPreference() {
```
jewzaam: Only read preference option is tested, but writeConcern and maxQueryTimeMS were included in implementation.

core/src/main/java/com/redhat/lightblue/client/Execution.java
```diff
@@ -0,0 +1,46 @@
...
+public class Execution extends Expression {
+
+    public static class MongoController {
+
+        public static final String READ_PREFERENCE = "readPreference",
+                WRITE_CONCERN = "writeConcern",
+                MAX_QUERY_TIME_MS = "maxQueryTimeMS";
+
+        public enum ReadPreference {
+            nearest, primaryPreferred, secondary, secondaryPreferred;
```
dcrissman: Is there not also a "primary" option?

core/src/main/java/com/redhat/lightblue/client/request/AbstractLightblueDataRequest.java
```diff
@@ -45,6 +49,15 @@ protected void appendRangeToJson(ObjectNode node, Integer begin, Integer maxResu
         }
     }
 
+    public void execution(Execution execution) {
+        this.execution = execution;
```
dcrissman: Just a comment really, but other methods on the data request objects return themselves for chaining purposes. Not a requirement, but might be nice.

dcrissman:
Is Execution needed on GenerateRequest or LiteralDataRequest?

dcrissman:
This might be a follow up change, but it seems like we would also want a way to specify these options in the lightblue-client.properties file. Maybe this could be the default for all requests, but it can be overridden by setting the Execution directly?

jewzaam:
@dcrissman +1 on clients being able to set a default.  Agree this is a follow up.  Could you log an issue if you haven't?

core/src/main/java/com/redhat/lightblue/client/request/data/DataDeleteRequest.java
paterczm: Range is not used for delete requests, so can't be set in AbstractLightblueDataRequest (common base).

paterczm:
@dcrissman 

> Is Execution needed on GenerateRequest or LiteralDataRequest?

The point of LiteralDataRequest is that you need to provide full request body yourself, including execution.

I guess writeConcern would make sense for GenerateRequest, but value generator implementation in lightblue-mongo does not support it. Same with locking functionality. For now, execution is only supported for standard crud operations.

core/src/main/java/com/redhat/lightblue/client/request/data/DataDeleteRequest.java
dcrissman: Perhaps we should spawn this discussion into it's own issue? That said, we could look at doing something like what we already do in lightblue-core and use interfaces... see https://github.com/lightblue-platform/lightblue-core/blob/master/crud/src/main/java/com/redhat/lightblue/crud/WithRange.java

dcrissman:
>> Is Execution needed on GenerateRequest or LiteralDataRequest?

>The point of LiteralDataRequest is that you need to provide full request body yourself, including execution.

>I guess writeConcern would make sense for GenerateRequest, but value generator implementation in lightblue-mongo does not support it. Same with locking functionality. For now, execution is only supported for standard crud operations.

Sorry, I was not arguing for them to have the fields. What I meant is that Generated/LiteralRequest both inherit from AbstractLightblueDataRequest, so by adding the getter/setter for Execution to AbstractLightblueDataRequest they both get them. So I meant to ask do we want these Requests to also have Execution?

core/src/main/java/com/redhat/lightblue/client/request/data/DataDeleteRequest.java
paterczm: This is problematic. AbstractLightblueDataRequest does not know if ArrayNode or ObjectNode is going to be used. This decision is made in child implementation, so I can't populate json body in AbstractLightblueDataRequest.

bserdar: You can reimplement the getBodyJson if an array is going to be used.

Is there a case an array will be used? I think it is all objects.

paterczm:
@dcrissman, I see what you mean now. You're right. I added one more abstract layer (AbstractLightblueDataExecutionRequest) so that GenerateRequest and LiteralDataRequest do not provide execution api. I don't like having so many abstract layers, but an alternative is some redundancy in crud requests, which I also don't like.

paterczm:
All feedback is incorporated now, unless otherwise noted in the comments.

core/src/main/java/com/redhat/lightblue/client/request/AbstractLightblueDataExecutionRequest.java
```diff
@@ -0,0 +1,28 @@
+package com.redhat.lightblue.client.request;
+
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.redhat.lightblue.client.Execution;
+
+public abstract class AbstractLightblueDataExecutionRequest extends AbstractLightblueDataRequest {
```
dcrissman: I still think an interface might be better. appendExecutionToJson could be a static method on the interface.

If the method "execution" returned AbstractLightblueDataRequest, then each implementation could return it's own class sub-type.

This just means each implementation would need to have it's own Execution instance, but that precedence has already been set with Projection and whatnot. We could go back and change appendRangeToJson to be the same in another PR.

core/src/main/java/com/redhat/lightblue/client/request/data/DataDeleteRequest.java
paterczm: > Is there a case an array will be used? I think it is all objects.

You're right, I checked incorrectly. Made the change. Didn't touch range, implementing it in a similar fashion would require one more abstract layer (because delete request does not offer range api) and it does not look like improvement over what we have.

core/src/main/java/com/redhat/lightblue/client/request/AbstractLightblueDataExecutionRequest.java
paterczm: Made appendExecutionToJson private. Execution bits are added in AbstractLightblueDataExecutionRequest.getBodyJson() now.

core/src/main/java/com/redhat/lightblue/client/request/AbstractLightblueDataExecutionRequest.java
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    private void appendExecutionToJson(ObjectNode node) {
+        if (execution != null)
+            node.set("execution", execution.toJson());
+    }
+
+    @Override
+    public JsonNode getBodyJson() {
+        ObjectNode node = JsonNodeFactory.instance.objectNode();
```
bserdar: Why not push creation of the object node to super, and only add the execution here?

core/src/main/java/com/redhat/lightblue/client/request/AbstractLightblueDataExecutionRequest.java
```diff
@@ -0,0 +1,37 @@
+package com.redhat.lightblue.client.request;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.JsonNodeFactory;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.redhat.lightblue.client.Execution;
+
+public abstract class AbstractLightblueDataExecutionRequest extends AbstractLightblueDataRequest {
```
bserdar: Name is a bit misleading. AbstractLightblueRequestWithExecution?

core/src/test/java/com/redhat/lightblue/client/request/TestDataUpdateRequest.java
```diff
@@ -46,4 +47,31 @@ public void testProjectionsAsArray() {
                 "\"projection\":{\"field\":\"*\",\"include\":true,\"recursive\":false}"));
     }
 
+    @Test
+    public void testExecutionReadPreference() {
+        DataUpdateRequest request = new DataUpdateRequest("fake");
+        request.execution(Execution.MongoController.withReadPreference(Execution.MongoController.ReadPreference.primaryPreferred));
```
dcrissman: @paterczm - I know this PR has already been merged, but is there a way we could short cut this call? Is MongoController even needed?

PR #299, opened 2016-05-17
fixes #289 - Add Literal consructor support for Array types

### LINKED ISSUE #289: Update $set with array value is very awkward
I could not find another way in the API other than this

```java
Update.set("survivorOfIds", Literal.value(Literal.toJson(Literal.values(survivorOfIds)))));
```

dcrissman:


core/src/main/java/com/redhat/lightblue/client/Literal.java
```diff
@@ -70,6 +70,25 @@ public static JsonNode valueToJson(Object x) {
...
+            }
+            else if (x instanceof Number[]) {
+                node = toJson(numbers((Number[]) x));
+            }
+            else if (x instanceof int[]){
+                node = toJson(values((int[]) x));
+            }
+            else if (x instanceof long[]){
+                node = toJson(values((long[]) x));
+            }
```
bserdar: Date[]
boolean[]
Boolean[]

dcrissman: Requested changes made

PR #300, opened 2016-05-17
fixes #298 - configure Execution globally

### LINKED ISSUE #298: Add ability for execution preferences to be globally configured in lightblue-client.properties
Spin off of #297

This might be a follow up change, but it seems like we would also want a way to specify these options in the lightblue-client.properties file. Maybe this could be the default for all requests, but it can be overridden by setting the Execution directly?

dcrissman:
Execution values can now globally be set in lightblue-client.properties, but can be overridden if set directly on the request.


http/src/main/java/com/redhat/lightblue/client/http/LightblueHttpClient.java
```diff
@@ -286,5 +296,37 @@ private static HttpTransport defaultHttpClientFromConfig(LightblueClientConfigur
             throw new RuntimeException(e);
         }
     }
+    
+    private Execution getExecutionInstance() {
+        if (execution == null) {
+            synchronized (this) {
+                if (execution == null) {
```
paterczm: Perhaps it would be easier if Execution constructor would be public (change needed) and you'd only use add* methods?

paterczm:
Seeing a lot of unrelated changes, refactorings probably. Could you make those at least in a separate commit?

http/src/main/java/com/redhat/lightblue/client/http/LightblueHttpClient.java
dcrissman: I have been debating that also, but I am not sure what an empty Execution would do. Use mongo defaults?

paterczm: Empty execution should be the same as no execution, so yes, whatever is configured in datasources.json.

dcrissman:
> Seeing a lot of unrelated changes, refactorings probably. Could you make those at least in a separate commit?

@paterczm - Just in LightblueClientConfiguration, and most of that is because I had my IDE generate the toString/equals/hashCode methods to include the new fields I was adding. Looks like some unnecessary "this." were accidently removed also, but everything else is part of this change.

paterczm:
Ok, I didn't know what happened with equals.

dcrissman:
Ready for review

core/src/main/java/com/redhat/lightblue/client/PropertiesLightblueClientConfiguration.java
```diff
@@ -57,6 +58,9 @@
     private static final String CERT_FILE_PATH_KEY = "certFilePath";
     private static final String CERT_PASSWORD_KEY = "certPassword";
     private static final String COMPRESSION = "compression";
+    private static final String READ_PREFERENCE = "readPreference";
+    private static final String WRITE_CONCERN = "writeConcern";
+    private static final String MAX_QUERY_TIME_MS = "maxQueryTimeMS";
```
paterczm: We already have those constants in MongoController, maybe use those?

paterczm: Talking nonsense, never mind.

http/src/main/java/com/redhat/lightblue/client/http/LightblueHttpClient.java
```diff
@@ -287,4 +298,39 @@ private static HttpTransport defaultHttpClientFromConfig(LightblueClientConfigur
...
+                        }
+                    }
+
+                    execution = mongoExecution;
+                }
+            }
+        }
+
+        return execution;
+    }
```
paterczm: This is unnecessarily complex. Let's simplify by allowing empty Execution creation. If you use package visibility, the constructor will be protected from misuse outside the client.

core/src/main/java/com/redhat/lightblue/client/PropertiesLightblueClientConfiguration.java
dcrissman: I changed these to be "execution.mongo.*" to help distinguish if we do add execution options for another backend.

http/src/main/java/com/redhat/lightblue/client/http/LightblueHttpClient.java
dcrissman: Simplified and moved to LightblueClientConfiguration so that it could be re-used and take advantage of visibility as you suggested.

dcrissman:
@paterczm - Ready for review

core/src/main/java/com/redhat/lightblue/client/LightblueClientConfiguration.java
```diff
@@ -32,6 +32,7 @@ public static Compression parseCompression(String str) {
     private ReadPreference readPreference;
     private String writeConcern;
     private Integer maxQueryTimeMS;
+    private transient Execution execution;
```
paterczm: Why transient?

dcrissman: No need for it to ever be serialized

paterczm: Why is it different in this regard than other LightblueClientConfiguration fields?

dcrissman: Maybe it is not. I just thought since it was composed from some of the other fields that there would be no need to ever serialize it. Not a big deal though, I can remove it if you would prefer?

paterczm: Just trying to understand how is it used.

PR #301, opened 2016-05-17
remove MongoController from Execution

dcrissman:


dcrissman:
@paterczm I am thinking we can just get rid of Execution.MongoController. Thoughts?

paterczm:
It's there to emphasize that those are mongo only options, not applicable to rdbms or ldap. Why do you want to remove it?

dcrissman:
> It's there to emphasize that those are mongo only options, not applicable to rdbms or ldap. Why do you want to remove it?

Ok, I can understand the argument that we might want different Executions for different backends. So with that in mind, then I question the add* methods that are on Execution. They would not be relevant to other backend types.

dcrissman:
@paterczm - What if I simply rename Execution in this PR to MongoExecution? Then we just plan to create an LdapExecution or MariaDbExecution if and when they are needed?

paterczm:
Makes sense. Let's make Execution and MongoExecution it's child.

dcrissman:
@paterczm - Hows this?

PR #310, opened 2016-08-24
Add support for specifying multiple CA certs in client configuration

### LINKED ISSUE #308: Allow client to specify more than one CA certificate for truststore

derek63:
Fixes #308


core/src/main/java/com/redhat/lightblue/client/LightblueClientConfiguration.java
```diff
@@ -54,6 +59,33 @@ public LightblueClientConfiguration(LightblueClientConfiguration configuration)
...
+        return (caFilePath.contains(CA_CERT_FILE_DELIMITER));
+    }
+
+    /**
+     *
+     * @return List<String> of caFilePaths generated from parsing the caFilePath
+     * value using the value specified by the CA_CERT_FILE_DELIMITER constant.
+     * If caFilePath value is not delimited, returns a List with only one path in it.
+     */
+    public List<String> getCaFilePaths() {
```
paterczm: How should getCaFilePath() behave when there are multiple CAs defined in the config file?

derek63: Ah good catch, IllegalStateException maybe?

PR #314, opened 2016-09-13
Add ability to check classpath for specified config file instead of o…

### LINKED ISSUE #313: Also check classpath when path to lightblue-client config file specified via constructor
The current code will only look for specified configuration files on the file system and not the classpath instead/also.

derek63:
…nly supporting file system.  Fixes #313.


pom.xml
```diff
@@ -145,7 +145,7 @@
     <sonar.projectName>lightblue-client</sonar.projectName>
     <rpm.install.basedir>/usr/share/jbossas/standalone/deployments</rpm.install.basedir>
 
-    <lightblue.rest.version>2.9.0-SNAPSHOT</lightblue.rest.version>
+    <lightblue.rest.version>2.9.0</lightblue.rest.version>
```
paterczm: I believe this should be 2.10.0-SNAPSHOT.

PR #322, opened 2016-11-04
Adding ordered flag to bulk request

paterczm:
For https://github.com/lightblue-platform/lightblue-core/pull/664


core/src/main/java/com/redhat/lightblue/client/request/DataBulkRequest.java
```diff
@@ -17,6 +17,8 @@
  */
 public class DataBulkRequest extends LightblueRequest {
 
+    private boolean ordered = true;
+
     protected List<CRUDRequest> requests=new ArrayList<>();
 
     public DataBulkRequest() {
```
bserdar: How about another ctor with ordered flag?

paterczm: Added.

PR #327, opened 2016-12-12
Add support for locking with POST

derek63:


core/src/main/java/com/redhat/lightblue/client/LightblueClient.java
```diff
@@ -12,6 +11,14 @@
 
     Locking getLocking(String domain);
 
+    /**
+     * This method allows
+     * @param domain
+     * @param usePost
+     * @return
+     */
+    Locking getLocking(String domain, Boolean usePost);
```
bserdar: I don't think there's a point to preserving the old lock API calls in the client. One ctor should be enough, using only POST.

http/src/main/java/com/redhat/lightblue/client/http/LightblueHttpClient.java
```diff
@@ -63,34 +106,39 @@ public LockingRequest(String domain, String callerId, String resourceId, Long tt
...
-            b.append(uri);
-            return b.toString();
+        private String getRestURIWithPost() {
+            return "lock/";
         }
     }
 
     private final class LockingImpl extends Locking {
 
+        Boolean usePost = false;
```
bserdar: Same here: no need for usePost. Client can always use POST.

http/src/main/java/com/redhat/lightblue/client/http/LightblueHttpClient.java
```diff
@@ -76,21 +80,19 @@ public String getRestURI(String baseServiceURI) {
             if (!baseServiceURI.endsWith("/")) {
                 b.append('/');
             }
-            b.append(uri);
+            b.append("lock/");
```
bserdar: trailing /?

derek63: I don't think its needed, but saw it elsewhere so probably just being consistent

derek63: Works with no / too, I will remove it

PR #329, opened 2017-01-12
Fixes #328: Allow injecting mapper used for request serialization

### LINKED ISSUE #328: Injected ObjectMapper not actually used for request serialization
Request serialization relies on getBody() implementation of request objects, and these delegate to the default object mapper.

So the only way to configure the object mapper used is to mutate the global default or set a new global default via the JSON class.

It looks like this is not a problem with responses.

Architecturally, requests are not created via the client (using it as a factory), but instead they're are newed up independently in client code, which means it's up to clients to inject the right ObjectMapper, and currently it's not injectable at all. One way would be to avoid the `String getBody()` method on requests, which implies a need for serialization, which implies a need to be able to customize serialization, and instead work exclusively from the `JsonNode getBodyJson()` method.

alechenninger:
#328 

TODO:

- [ ] Does removing the content length header from request prevent persistent connections?

### DIFF
--- a/core/src/main/java/com/redhat/lightblue/client/request/LightblueRequest.java
+++ b/core/src/main/java/com/redhat/lightblue/client/request/LightblueRequest.java
@@ -39,6 +39,7 @@ public abstract class LightblueRequest implements Serializable {
     /**
      * Requst body as string, defaults to getBodyJson().toString()
      */
+    @Deprecated
     public String getBody() {
         JsonNode body=getBodyJson();
         if(body!=null) {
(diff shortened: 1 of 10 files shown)


core/src/main/java/com/redhat/lightblue/client/request/LightblueRequest.java
```diff
@@ -39,6 +39,7 @@ public HttpMethod getHttpMethod() {
     /**
      * Requst body as string, defaults to getBodyJson().toString()
      */
+    @Deprecated
```
dcrissman: Why? Or perhaps what should be used instead?

alechenninger: getBodyJson should be used instead. This uses a default global object mapper so it can't be configured. I'll add a comment.

alechenninger: Well, it can be configured, but isn't configured the same as for responses. Details in #328

bvulaj:
@alechenninger @dcrissman  Should we have merged this forever ago?

PR #332, opened 2017-01-30
Add support for passing result metadata, and updateIfCurrent flag

bserdar:


core/src/main/java/com/redhat/lightblue/client/request/data/DataSaveRequest.java
```diff
@@ -41,7 +47,7 @@ public DataSaveRequest returns(List<? extends Projection> projection, Integer be
         return this;
     }
 
-    public DataSaveRequest returns(Projection... projection) {
+   public DataSaveRequest returns(Projection... projection) {
```
paterczm: Whitespace introduced.

core/src/main/java/com/redhat/lightblue/client/response/DefaultLightblueDataResponse.java
```diff
@@ -52,6 +52,25 @@ public int parseMatchCount() {
...
+     * array.
+     */
+    @Override
+    public ResultMetadata[] getResultMetadata() throws LightblueParseException {
+        JsonNode node=getJson().get("resultMetadata");
+        if(node instanceof ArrayNode) {
+            try {
+                return getMapper().readValue(node.traverse(),ResultMetadata[].class);
+            } catch (IOException e) {
+                throw new LightblueParseException("Error parsing lightblue response:"+getText()+"\n",e);
```
paterczm: This will only happen when this client is used against older Lightblue server version (from before update if current) - right?

bserdar:
It this client is used with older server, getResultMetadata will return
null. The exception will only happen with a malformed response.

PR #334, opened 2017-04-28
Streaming results

bserdar:


http/src/main/java/com/redhat/lightblue/client/http/LightblueHttpClient.java
```diff
@@ -268,6 +276,101 @@ public void close() throws IOException {
...
+                            }
+                        };
+                    streamParser.parse();
+                    
+                } catch (IOException e) {
+                    throw new LightblueException(e);
+                } finally {
+                    try {
+                        stream.close();
+                    } catch (Exception e) {}
```
bvulaj: Worth logging something?

bserdar:
Added logging

alechenninger:
AFAIK you should be able to stream the parsing of the HTTP response regardless of whether lightblue is streaming results as it iterates over the cursor or as it serializes to JSON. To the client, it's chunked encoding JSON either way.

bserdar:
Is that really sent as chunked-encoding? I think I got responses with fixed
content-lengths (not chunked-encoding).

PR #336, opened 2017-06-06
Adding writeConcern enum

paterczm:
This is to match how readPreference is used.


core/src/main/java/com/redhat/lightblue/client/Expression.java
```diff
@@ -29,6 +29,7 @@ protected Expression(boolean arrayNode) {
     /**
      * Add field:value
      */
+    // TODO: add is using set, this is confusing. Rename all adds to sets?
```
dcrissman: open an issue?

PR #337, opened 2017-07-26
Added lightblue health check feature

sameer31:
Added lightblue health check feature to be used by services to check lightblue health
@alechenninger, @svkm102, @macgregor please review


core/src/main/java/com/redhat/lightblue/client/request/LightblueHealthRequest.java
```diff
@@ -0,0 +1,41 @@
...
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.redhat.lightblue.client.http.HttpMethod;
+
+public class LightblueHealthRequest extends LightblueRequest{
+
+    private static final long serialVersionUID = 2479314164852663290L;
+    
+    private String diagnosticsUri;
+
```
svkm102: maybe use healthURL for consistency and clarity.

core/src/main/java/com/redhat/lightblue/client/response/LightblueHealthResponse.java
```diff
@@ -0,0 +1,83 @@
...
+import com.redhat.lightblue.client.LightblueException;
+
+/**
+ * @author ssaurabh
+ *
+ */
+/**
+ * @author ssaurabh
+ *
+ */
```
svkm102: multiple @author

hystrix/src/main/java/com/redhat/lightblue/client/hystrix/LightblueHystrixClient.java
```diff
@@ -252,4 +254,10 @@ public ResultStream prepareFind(DataFindRequest req) throws LightblueException {
         return new ResultStream(new StreamingClosure(req),null);
     }
 
+    @Override
+    public LightblueHealthResponse lightblueHealth(LightblueHealthRequest lightblueHealthRequest)
+            throws LightblueException {
+        return null;
+    }
+
 }
```
svkm102: For consistency purpose, it would be nice to have a proper hystrix implementation of this method.

hystrix/src/test/java/com/redhat/lightblue/client/hystrix/LightblueHystrixClientTest.java
```diff
@@ -67,6 +69,12 @@ public ResultStream prepareFind(DataFindRequest req) throws LightblueException {
             return null;
         }
 
+        @Override
+        public LightblueHealthResponse lightblueHealth(LightblueHealthRequest lightblueHealthRequest)
+                throws LightblueException {
+            return null;
+        }
+
     }
```
svkm102: same as above.

core/src/main/java/com/redhat/lightblue/client/request/LightblueHealthRequest.java
```diff
@@ -0,0 +1,41 @@
+package com.redhat.lightblue.client.request;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.redhat.lightblue.client.http.HttpMethod;
+
+public class LightblueHealthRequest extends LightblueRequest{
```
alechenninger: Is this intended to be for /health or /diagnostics?

These are two distinct APIs with different response schemas. The (future) intention of /health is to return an empty response with just a status code 200 or 503. We may add a response body here in the future but the primary use case is a boolean "should you send me requests?" answer.

The /diagnostics endpoint is different, it's a collection of various health checks and details of their results and potentially other useful information. This has a stable schema.

I would have support for both, though right now we only have /diagnostics implemented.

core/src/main/java/com/redhat/lightblue/client/request/LightblueHealthRequest.java
```diff
@@ -0,0 +1,41 @@
...
+    
+    private String diagnosticsUri;
+
+    public LightblueHealthRequest() {
+        this("diagnostics");
+    }
+    
+    public LightblueHealthRequest(String diagnosticsUri) {
+        super(HttpMethod.GET);
+        this.diagnosticsUri = diagnosticsUri;
```
alechenninger: It doesn't make sense to parameterize the path (it's not a URI) for either health or diagnostics because they are a fixed part of lightblue's public API, just like /find or /insert, etc.

core/src/main/java/com/redhat/lightblue/client/response/LightblueHealthResponse.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    public boolean isHealthy() {
+        return parseHealthy();
+    }
+
+    
+    /**
+     * parses the lightblue health response and returns true only if healthy is
+     * true for all the nodes in the response, otherwise returns false
```
alechenninger: Going back to the /health vs /diagnostics distinction, this looks like you are talking about the /diagnostics endpoint (which makes sense, because this is the only one that exists right now).

For /diagnostics, we should be careful encoding logic in the client that says what "healthy" means. It depends on use case and lightblue server configuration.  I would probably consider not providing this kind of API in the client, and instead have a class that encapsulates the response and allows the application using the client to easily examine health of controllers, etc, that may be included in the diagnostics response. I forget what the schema of the response is but basically it should follow that.

http/src/main/java/com/redhat/lightblue/client/http/LightblueHttpClient.java
```diff
@@ -405,4 +407,17 @@ private static HttpTransport defaultHttpClientFromConfig(LightblueClientConfigur
         }
     }
 
+    @Override
+    public LightblueHealthResponse lightblueHealth(LightblueHealthRequest lightblueHealthRequest)
```
alechenninger: lightblue is probably redundant here. Maybe just "diagnostics()"?

There's probably not a need to pass a request object for diagnostics or health either because they don't have any parameters.

http/src/test/java/com/redhat/lightblue/client/http/LightblueHttpClientTest.java
```diff
@@ -273,6 +275,50 @@ public boolean processDocument(ResultStream.StreamDoc doc) {
...
+            + "\"message\":\"LDAPConnection [DN: uid=lightblueapp,ou=serviceusers,ou=lightblue,dc=redhat,dc=com, "
+            + "Status: true]\"}}";
+   
+    @Test
+    public void testLightblueDiagnosticsElementPresent() throws Exception {
+        
+        DiagnosticsElement expectedElementDiagnostics = new DiagnosticsElement("MongoCRUDController", true, "[Mongo Config [lightbluemongo1.dev.a1.vary.redhat.com:27017, "
+                + "DatabaseName: metadata]=>ping:OK, Mongo Config [lightbluemongo1.dev.a1.vary.redhat.com:27017, DatabaseName: data]=>ping:OK]");
+        
+        when(httpTransport.executeRequest(any(LightblueRequest.class), anyString())).thenReturn(new FakeResponse(diagnosticsResponse, null));
```
alechenninger: When using mocks I think it's a good idea to make sure the right parameters are being passed as well. So for example we should probably test the base URI is correct.

We also need a unit test for the diagnostics request itself. There are others like it in lightblue-client/core/src/test/java/com/redhat/lightblue/client/request .

http/src/test/java/com/redhat/lightblue/client/http/LightblueHttpClientTest.java
```diff
@@ -273,6 +275,50 @@ public boolean processDocument(ResultStream.StreamDoc doc) {
...
+            + "Mongo Config [lightbluemongo1.dev.a1.vary.redhat.com:27017, DatabaseName: data]=>ping:OK]\"},"
+            + "\"ldap-auth-healthcheck\":{\"healthy\":true,"
+            + "\"message\":\"LDAPConnection [DN: uid=lightblueapp,ou=serviceusers,ou=lightblue,dc=redhat,dc=com, "
+            + "Status: true]\"}}";
+   
+    @Test
+    public void testLightblueDiagnosticsElementPresent() throws Exception {
+        
+        DiagnosticsElement expectedElementDiagnostics = new DiagnosticsElement("MongoCRUDController", true, "[Mongo Config [lightbluemongo1.dev.a1.vary.redhat.com:27017, "
+                + "DatabaseName: metadata]=>ping:OK, Mongo Config [lightbluemongo1.dev.a1.vary.redhat.com:27017, DatabaseName: data]=>ping:OK]");
```
alechenninger: Please do not expose URLs publicly.

core/src/main/java/com/redhat/lightblue/client/response/LightblueDiagnosticsResponse.java
```diff
@@ -0,0 +1,6 @@
+package com.redhat.lightblue.client.response;
+
+public interface LightblueDiagnosticsResponse extends LightblueResponse {
+
+    DiagnosticsElement getDiagnostics(String diagnosticsElementName);
```
alechenninger: We should also add an API to get all of the diagnostics elements.

core/src/main/java/com/redhat/lightblue/client/response/DefaultLightblueDiagnosticsResponse.java
```diff
@@ -61,27 +63,44 @@ public boolean hasDiagnostics(String diagnosticsElementName) {
     public List<DiagnosticsElement> getDiagnostics() {
         List<DiagnosticsElement> diagnosticsElements = new ArrayList<>();
 
-        Iterator<String> fieldNames = getJson().fieldNames();
-        boolean isHealthy = false;
-        String message = null;
+        if (getJson() != null) {
+            Iterator<String> fieldNames = getJson().fieldNames();
+            boolean isHealthy = false;
+            String message = null;
```
alechenninger: Still a bit weird that these are declared outside of the `while` block below. They are only needed within that block, so being declared in an outer scope makes it look like they are used in the outer scope when they aren't.

Not a huge deal.

core/src/test/java/com/redhat/lightblue/client/request/TestLightblueDiagnosticsResquest.java
```diff
@@ -11,5 +11,10 @@
     public void testGetRestURI() {
         Assert.assertEquals(finalDiagnosticsURI, testRequest.getRestURI(baseURI));
     }
+    
+    @Test
+    public void testRequestBodyNull() {
+        Assert.assertEquals(null, testRequest.getBodyJson());
+    }
 
 }
```
alechenninger: Nitpick: typo in this class name (Request instead of Resquest)

PR #338, opened 2017-07-31
Added SystemEnvironmentAwareProperties

ykoer:
to support system environment placeholders in the lightblue client configuration.

The rationale behind it is Openshift deployment. Configurations are usually passed in as System Environment variables.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -7,4 +7,17 @@
   </parent>
   <version>5.12.0-SNAPSHOT</version><artifactId>lightblue-client-core</artifactId>
   <name>lightblue-client: ${project.groupId}|${project.artifactId}</name>
+
+  <dependencies>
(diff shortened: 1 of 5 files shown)


core/pom.xml
```diff
@@ -7,4 +7,17 @@
...
   <version>5.12.0-SNAPSHOT</version><artifactId>lightblue-client-core</artifactId>
   <name>lightblue-client: ${project.groupId}|${project.artifactId}</name>
-</project>
\ No newline at end of file
+
+  <dependencies>
+    <dependency>
+      <groupId>org.powermock</groupId>
+      <artifactId>powermock-module-junit4</artifactId>
+      <version>1.7.0</version>
```
bserdar: scope=test please, for both

